PR #1, opened 2020-09-30
add logo

rohieb:


rohieb:
Should've rendered the text to paths…

fscherf:
Thanks for that! \o/

PR #3, opened 2021-08-11
Fix SyntaxError: invalid syntax

scaleoutsean:


fscherf (commented):
That changes the order of the routes. Adding a "," to the first route should fix the problem.

scaleoutsean:
OK feel free to made edits as you see fit!

PR #5, opened 2021-08-13
Fix single-file script examples

laundmo:
example broke, now example less broke


fscherf:
Ugh! Stupid tab completion mistake. Thanks! I currently have no PC on hand, will merge as soon as possible

fscherf:
@laundmo Thanks!

PR #10, opened 2021-08-15
Use `del` keyword to remove attribute

### LINKED ISSUE #9: Failed to enable button
An exception is raised when I try to enable disabled button.
```
    File "/Users/marat.reymers/.local/share/virtualenvs/blot-4lfKRN_4/lib/python3.9/site-packages/lona/html/nodes.py", line 176, in disabled
      self.attributes.remove('disabled')
  AttributeError: 'AttributeDict' object has no attribute 'remove'
```

It seems need to use `pop` method or `del` keyword here https://github.com/fscherf/lona/blob/7fd5f10df8f46b0f1dd5dc643d871467d45474a8/lona/html/nodes.py#L176

maratori:
Fixes #9.
There is no such method `remove` in attributes.

### DIFF
--- a/lona/html/nodes.py
+++ b/lona/html/nodes.py
@@ -173,7 +173,7 @@ class Button(Node):
                     self.attributes['disabled'] = 'True'
             else:
                 if 'disabled' in self.attributes:
-                    self.attributes.remove('disabled')
+                    del self.attributes['disabled']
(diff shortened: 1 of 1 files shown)


fscherf:
Great! :) thanks for contributing!

PR #13, opened 2021-08-20
Correct way of linking to other pages

### LINKED ISSUE #12: Documentation links are broken
Cross-reference links in the documentation are pointing to `.rst` pages, where they should be pointing to `.html`.

See for example page https://lona-web.org/end-user-documentation/frontends.html. It has a link to 

_More information:_ [Basic Concept](https://lona-web.org/end-user-documentation/basic-concept.rst)

Defined in:

https://github.com/lona-web-org/lona/blame/master/doc/content/end-user-documentation/frontends.rst#L12

Either documentation has to switch from `.rst` to `.html` in links, or building of the documentation has be fixed.

if someone can advice which approach should be taken, I can have a closer look.

korantu:
Reference - https://flamingo-web.org/user/writing_content.html

fixes #12

### DIFF
--- a/doc/content/end-user-documentation/frontends.rst
+++ b/doc/content/end-user-documentation/frontends.rst
@@ -9,7 +9,8 @@ is supposed to load the Lona Javascript client, define the basic layout of the
 website and load all static files.
 
 **More information:**
-`Basic Concept </end-user-documentation/basic-concept.rst>`_
+`Basic Concept </end-user-documentation/basic-concept.html>`_
+
 
 Lona has a default frontend view build in that uses an overrideable template.
 To customize your Lona application you can override the template or the entire
@@ -160,7 +161,8 @@ On the server all messages get handled by middlewares.
 
             print(message)
 
-**More information:** `Middlewares </end-user-documentation/middlewares.rst>`_
+**More information:**
+`Middlewares </end-user-documentation/middlewares.html>`_
 
 
 Server To Client


fscherf:
Hi @korantu,

I tested your patch and what i said in #12 does not apply here (my bad, not yours). Flamingo links are based on jinja2 templating and the frontends page shows templates. To stop Flamingo from rendering the examples ``is_template: False`` is set. Therefore the link syntax gets uninterpreted.
You have to use rst syntax in those cases.

To test the docs you can go to ``lona/doc`` and run ``make server``.

korantu:
Indeed, this one page is not a template

https://github.com/lona-web-org/lona/blame/master/doc/content/end-user-documentation/frontends.rst#L1

@fscherf , thanks for the documentation test hint, this way it works.

Why is this setting only on this page?

fscherf:
@korantu: In Flamingo any content file can be a Jinja2 template to make syntax like
```text
{{ link('link text', 'link-target.rst') }}
```
or mixed syntax like

```text
Hello {{ variable.name }}
=========================

{% for i in range(10) %}
    Hello World
    -----------
{% endfor %}

```

work. If you want to add examples of Jinja2 templates to your content file, and you want Flamingo to not render them you have to disable templating for this file.
This is the only file that defines Jinja2 examples, thats why thats the only file with this setting disabled.

fscherf:
@korantu: I rebuilt the docs and deployed them to lona-web.org. Thanks for contributing! :)

PR #21, opened 2021-08-22
Redefine `input_event.node` in `handle_input_event`

### LINKED ISSUE #20: Wrong `event.node` for `CheckBox`
When `CheckBox` fires event, it has wrong attribute `event.node`. I expect to see `CheckBox` itself, but receives `CheckBox.input_node`. Here is an example with workaround.

```py
from lona import LonaApp
from lona.html import CheckBox
from lona.view import LonaView

app = LonaApp(__file__)

@app.route("/")
class CheckBoxView(LonaView):
    def __init__(self, server, view_runtime, request):
        super().__init__(server, view_runtime, request)
        self.checkbox = CheckBox(bubble_up=True)

    def handle_request(self, request):
        return self.checkbox

    def handle_input_event(self, event):
        if event.node == self.checkbox:
            print("this doesn't work")
        if event.node == self.checkbox.input_node:
            print("workaround works")

app.run(port=8080)
```

It seems we need to change `input_event.node` here before return
https://github.com/lona-web-org/lona/blob/a9bd343a347508fa2b9069227b6d2de50fefd881/lona/html/data_binding/inputs.py#L107

maratori:
Fixes #20.
Users of `TextInput` and successors (`TextArea`, `CheckBox`) expect to see `input_event.node` to be `TextInput` itself, not its attribute `.input_node`.
The same if for `Select`.
So need to update it before bubble up.

### DIFF
--- a/doc/content/end-user-documentation/html.rst
+++ b/doc/content/end-user-documentation/html.rst
@@ -334,7 +334,7 @@ Select
 
     from lona.html import Select
 
-    CheckBox([
+    Select([
         # value, label, is_selected
         ('foo', 'Foo', True),
         ('bar', 'Bar', False),
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -104,6 +104,8 @@ class TextInput(Widget):
         self.value = input_event.data
 
         if self.bubble_up:
+            if input_event.node is self.input_node:
+                input_event.node = self
             return input_event
 
     # properties ##############################################################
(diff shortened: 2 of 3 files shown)


fscherf:
@maratori good idea. We should use ``is`` instead of ``==``. Because of testing reasons you can two different nodes with equal attributes with ``==``, so it could result in fals positives.

Can you add more description to your patch?

fscherf:
@maratori this seems to be broken in ``Select`` too. Can you apply this fix there too?

maratori:
@fscherf sure. I'm on it.

maratori:
@fscherf done

fscherf:
@maratori: I tested your changes with the test_project and everything seems to work. This make much more sense now! Thanks for contributing.
I will release 1.3 with your changes, and my changes of the weekend, tomorrow morning.

Message me if you want some stickers :)
https://lona-web.org/stickers.html

PR #34, opened 2021-08-30
fscherf/topic/html-api-improvements

fscherf:


fscherf:
@maratori, @laundmo, @korantu, @themutt: Has someone time to test this?

lona/html/data_binding/select.py
```diff
@@ -1,81 +1,23 @@
...
-        return {
-            'id': id_list,
-            'class': class_list,
-            'style': style,
-            **attributes,
-            **misc_kwargs,
-        }
+        self.disabled = disabled
 
     def handle_input_event(self, input_event):
```
maratori, line 22:
```suggestion
    def handle_change(self, input_event):
```

lona/html/data_binding/select.py
```diff
@@ -1,81 +1,23 @@
...
-    STYLE = {}
-    ATTRIBUTES = {}
+class Option(Node):
+    TAG_NAME = 'option'
 
-    def __init__(self, values=[], disabled=False, bubble_up=False,
-                 **select_kwargs):
 
-        self.bubble_up = bubble_up
+class Select(Node):
```
maratori, line 9: Don't you want to move it to `lona/html/nodes.py`?

lona/html/data_binding/inputs.py
```diff
@@ -1,101 +1,35 @@
...
         'type': 'text',
     }
 
-    def __init__(self, value=None, disabled=False, bubble_up=False,
-                 input_delay=DEFAULT_INPUT_DELAY, **input_kwargs):
-
-        self.bubble_up = bubble_up
-
-        self.input_node = self.gen_input_node(**input_kwargs)
+    def __init__(self, *args, value=None, disabled=False, bubble_up=False,
```
maratori, line 15: Input is self closing and can't contain children.
```suggestion
    def __init__(self, value=None, disabled=False, bubble_up=False,
```

lona/html/data_binding/inputs.py
```diff
@@ -1,101 +1,35 @@
...
-            'class': class_list,
-            'style': style,
-            **attributes,
-            **misc_kwargs,
-        }
-
-    def gen_input_node(self, **kwargs):
-        return self.NODE_CLASS(**self.gen_node_args(**kwargs))
-
     def handle_input_event(self, input_event):
```
maratori, line 28:
```suggestion
    def handle_change(self, input_event):
```

lona/html/data_binding/inputs.py
```diff
@@ -1,101 +1,35 @@
...
         if input_event.name != 'change':
             return input_event
 
-        self.input_node.attributes.__setitem__(
+        self.attributes.__setitem__(
             'value',
             input_event.data,
             issuer=(input_event.connection, input_event.window_id),
```
maratori, line 35: nit: Let's add explaining comment (looks like some hack)

lona/html/data_binding/inputs.py
```diff
@@ -104,101 +38,42 @@ def handle_input_event(self, input_event):
...
-        return self.input_node.has_id
-
-    @property
-    def has_class(self):
-        return self.input_node.has_class
+            elif 'disabled' in self.attributes:
+                self.attributes.pop('disabled')
 
 
 class TextArea(TextInput):
```
maratori, line 68: I believe `TextArea` is not `TextInput`.
```suggestion
class TextArea(Node):
```
For example now there is a problem: `TextInput` has to have `*args` in constructor just because `TextArea` is not self closing.

maratori (commented):
@fscherf I like this PR in general 👍 
But there are some comments from my side.

lona/html/data_binding/select.py
fscherf: Select is no widget anymore, but it keeps implementing data-binding. Therefore i would let it stay in ``lona.html.data_binding``.

lona/html/document.py
```diff
@@ -21,42 +19,22 @@ def add_patch(self, *args, **kwargs):
...
+        node = None
+        nodes = []
 
         with self.lock:
-            iter_nodes(self.html)
+            for _node in self.html.iter_nodes():
+                if _node.id == node_id:
+                    node = _node
+
+                    break
```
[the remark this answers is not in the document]
maratori: Without explicit exception there will be an implicit one on the next line
```
AttributeError: 'NoneType' object has no attribute 'parent'
```

lona/html/data_binding/select.py
maratori: What does it mean "data-binding"? It is not clear for me.

lona/html/data_binding/inputs.py
fscherf: TextArea has to inherit from TextInput to make the databinding aspect working. Also these two classes do pretty much the same thing. ``SELF_CLOSING_TAG`` for example gets only used by the backend when rendering an ``repr()`` string. The javascript based rendering engine does not use these informations.

lona/html/data_binding/select.py
fscherf: If you change the value of a ``TextInput`` in your browser, the data gets changed in the frontend and the backend. If you change it using Python ``text_input.value = 'foo'``, it also changes in the frontend and the backend. Thats called data-binding.

lona/events/input_event.py
```diff
@@ -41,10 +41,8 @@ def __init__(self, request, payload, document, connection, window_id):
         # find node
         # node info contains a lona node id
         if self.node_info[0] or self.node_info[1]:
-            self.node, self.widgets = document.get_node(
```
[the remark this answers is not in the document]
fscherf: This pr will have breaking changes anyway. Also ``InputEvent.widgets`` was never documented. Therefore i removed it.

lona/events/input_event.py
```diff
@@ -41,10 +41,8 @@ def __init__(self, request, payload, document, connection, window_id):
...
-                widget_id=self.node_info[0],
-                node_id=self.node_info[1],
-            )
+            self.nodes = document.get_node(node_id=self.node_info[1])
```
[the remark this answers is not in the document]
fscherf: done

lona/html/parsing.py
```diff
@@ -21,10 +28,83 @@
...
+
+    _find_node_classes(html)
+
+
 class NodeHTMLParser(HTMLParser):
+    def __init__(self, *args, use_high_level_nodes=True, node_classes={},
+                 **kwargs):
+
+        self.use_high_level_nodes = use_high_level_nodes
+        self.node_classes = node_classes
```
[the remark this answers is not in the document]
fscherf: Good point! Done

lona/html/document.py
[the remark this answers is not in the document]
fscherf: I see. Then we should add

``python
if not node:
    return []
```

in line 33

File: lona/html/data_binding/select.py (hunk 1 above)
Comment (fscherf) [reply lona/html/data_binding/select.py]: Done

File: lona/html/data_binding/inputs.py (hunk 3 above)
Comment (fscherf) [reply lona/html/data_binding/inputs.py]: Done

File: lona/html/data_binding/inputs.py (hunk 4 above)
Comment (fscherf) [reply lona/html/data_binding/inputs.py]: Done

File: lona/html/data_binding/inputs.py (hunk 5 above)
Comment (fscherf) [reply lona/html/data_binding/inputs.py:44]: I added the commit message of the original commit. Hm it could look like a hack because i use ``__setitem__`` directly. Maybe we should move that functionality to another function. Maybe not in this pr though.

**Comment by fscherf** [comment]
@maratori: That ``input`` cant be overridden by ``node_classes={}`` and we have to have a list of their classes in the parser is definitely unpractical. I will add a second constant ``INPUT_CLASSES`` where ``type`` links to specific classes. They than can be overridden by passing ``node_classes`` as a list, which gets to be split into inputs and normal nodes too.

File: lona/html/data_binding/select.py (hunk 2 above)
Comment (maratori) [reply lona/html/data_binding/select.py:9]: Am I right, that it is not enough to execute `text_input.value = 'foo'` on the server side? I also have to call `view.show()` to send patches to the browser.

File: lona/events/input_event.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (maratori) [reply lona/events/input_event.py:44]: I don't have permissions to resolve conversations. Do it for me, please.

File: lona/html/data_binding/select.py (hunk 2 above)
Comment (fscherf) [reply lona/html/data_binding/select.py:9]: Yes thats right

Comment (maratori) [reply lona/html/data_binding/select.py:9]: Seems now I've got the point.
Any `Node` has one way "binding" (changes from server goes to the browser). For example `node.class_list.add("foo")`.
But only nodes in `data_binding` package have two way binding for some attributes (for now it is `.value`).

File: lona/html/parsing.py [hunk 11]
```diff
@@ -21,10 +28,83 @@
...
 
+    def get_node_class(self, tag_name, attributes):
+        if not self.use_high_level_nodes:
+            return Node
+
+        # inputs
+        if tag_name == 'input':
+            input_type = attributes.get('type', 'text')
+
+            if input_type == 'text':
```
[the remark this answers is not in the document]
fscherf: done

lona/html/parsing.py
```diff
@@ -21,10 +28,83 @@
...
+
+
+def _setup_node_classes_cache():
+    from lona import html
+
+    _find_node_classes(html)
+
+
 class NodeHTMLParser(HTMLParser):
+    def __init__(self, *args, use_high_level_nodes=True, node_classes={},
```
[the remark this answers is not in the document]
fscherf: fixed

lona/html/data_binding/select.py
```diff
@@ -1,106 +1,44 @@
...
+    TAG_NAME = 'select'
+    EVENTS = [CHANGE]
 
-        self.select_node = SelectNode(**self.gen_node_args(**select_kwargs))
-        self.select_node.events.add(CHANGE)
+    def __init__(self, *args, values=[], disabled=False, bubble_up=False,
```
maratori, line 13: I didn't tried, but seems following will not work as expected
```py
HTML("<select><option value='a'>a</option><option value='b'>b</option></select>")
```
```py
HTML(Select(Option("a", value="a"), Option("b", value="b")))
```

fscherf: Oh! You are right! The ``Select`` node relies on getting its values as a list of strings, not ``Option`` objects. I will fix that

lona/html/data_binding/inputs.py
```diff
@@ -104,101 +47,42 @@ def handle_input_event(self, input_event):
...
-    NODE_CLASS = TextAreaNode
+    TAG_NAME = 'textarea'
+    SELF_CLOSING_TAG = False
```
maratori, line 79: nit:
```suggestion
    SELF_CLOSING_TAG = False
    ATTRIBUTES = {}
```

fscherf: fixed

lona/html/data_binding/select.py
fscherf: @maratori: Fixed!

lona/html/parsing.py
```diff
@@ -21,10 +28,88 @@
...
+
+            # nodes
+            else:
+                if node_class.TAG_NAME in NODE_CLASSES:
+                    logger.warning(
+                        'WARNING: Two Node classes with the same tag name were found: %s',  # NOQA
+                        node_class.TAG_NAME,
+                    )
+
+            NODE_CLASSES[node_class.TAG_NAME] = node_class
```
maratori, line 62: nit: This line will be executed multiple times with same tag name `input` but different classes. Is it by intention?
```suggestion
                NODE_CLASSES[node_class.TAG_NAME] = node_class
```

maratori (approved):

lona/html/parsing.py
fscherf: ACK! I fixed that

maratori (approved):
LGTM 👍

fscherf:
@maratori: Great! Thanks for all that review! If you want me to send you some stickers, email me
https://www.reddit.com/r/lona_web_org/comments/p5i7ch/we_got_stickers/

PR #39, opened 2021-09-01
Use immutable default values

maratori:


lona/server.py
```diff
@@ -447,9 +447,9 @@ async def _handle_http_request(self, http_request):
             return response
 
         # websocket requests ##################################################
-        if(http_request.method == 'GET' and
-           http_request.headers.get('upgrade', '').lower() == 'websocket'):
-
+        if (http_request.method == 'GET' and
+                http_request.headers.get('upgrade', '').lower() ==
+                'websocket'):
```
fscherf, line 452: This seems to be only a style change. Can you put this in another commit?

lona/view.py
```diff
@@ -37,9 +37,8 @@ def _add_view_to_objects(cls, view_object):
     def _remove_view_from_objects(cls, view_object):
         objects_key = cls._get_objects_key(view_object, is_class=False)
 
-        if(objects_key in cls._objects and
-           view_object in cls._objects[objects_key]):
-
+        if (objects_key in cls._objects and
+                view_object in cls._objects[objects_key]):
```
fscherf, line 41: This is also only a style change

fscherf (commented):
Besides some nitpicking great PR!

lona/server.py
maratori: Removed this change. It was done by my IDE, not intentionally.

lona/view.py
maratori: Removed this change. It was done by my IDE, not intentionally.

fscherf:
Great! Thanks! That looked like some tedious work

PR #41, opened 2021-09-01
Avoid possible `NameError`

maratori:


fscherf:
@maratori: does this change the default output color to green?

maratori:
@fscherf Yes

fscherf:
@maratori the default foreground color of the terminal should not be changed to make all sorts of terminal color palettes work. We recently had issues with light terminals.

maratori:
@fscherf I can change it, if you want.
Just wanted to note that I used the same config as for debug.

fscherf:
@maratori: No worries! I think we should initialize all values as empty strings.

When i fixed the color issue initially, a read a bunch of best practice articles and the common sense was, that foreground and background of the default output should remain unchanged. Coloring errors red works in most terminals and if debug does not work as expected, thats acceptable because thats only for developers or admins.

maratori:
@fscherf Done

fscherf:
@maratori: Great! Thanks :)

PR #42, opened 2021-09-01
Remove unused variables

maratori:


lona/routing.py
```diff
@@ -179,15 +179,11 @@ def _reverse(self, name, *args, **kwargs):
...
-        key_error = None
-
         try:
             return route.format_string.format(*args, **kwargs)
 
         except KeyError as e:
-            key_error = e
-
-        raise ValueError('missing URL arg: {} '.format(key_error.args[0]))
+            raise ValueError('missing URL arg: {} '.format(e.args[0]))
```
fscherf, line 186: This is actually on purpose to avoid stacking tracebacks

fscherf (changes requested):
Great! I am ok with all changes except for one.

lona/routing.py
maratori: Reverted and added comment

fscherf:
Thanks!

For the future: It would be nice if you could split up these kinds fixes in smaller commits with commit messages for better orientation like ``imports: acquire: remove unused variable``.

PR #43, opened 2021-09-01
Remove old methods `Node._xxx_patches`

maratori:
If I'm right, they are not used anymore (after 83b35ff42ef49513ba41f12c4489a108d3365f89).


fscherf:
@maratori: You are right. Thanks for the patch!

PR #44, opened 2021-09-01
Align behaviour of `AttributeDict` and `AttributeList`

maratori:
`AttributeList.remove()` didn't raise exception if `attribute` not found in list. However `AttributeDict.pop()` did raise exception if `name` not found in dict.
Now they behave the same - no exception.

### DIFF
--- a/lona/html/attribute_dict.py
+++ b/lona/html/attribute_dict.py
@@ -19,6 +19,9 @@ class AttributeDict:
 
     def pop(self, name):
         with self._node.lock:
+            if name not in self._attributes:
+                return None
+
             attribute = self._attributes.pop(name)
 
             self._node.document.add_patch(
@@ -91,17 +94,7 @@ class AttributeDict:
             )
 
     def __delitem__(self, name):
-        with self._node.lock:
-            del self._attributes[name]
-
-            self._node.document.add_patch(
-                node_id=self._node.id,
-                patch_type=self.PATCH_TYPE,
-                operation=OPERATION.REMOVE,
-                payload=[
-                    name,
-                ],
-            )
+        self.pop(name)
 
     def __eq__(self, other):
         with self._node.lock:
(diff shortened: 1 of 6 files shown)


lona/html/attribute_dict.py
```diff
@@ -19,6 +19,9 @@ def items(self):
 
     def pop(self, name):
         with self._node.lock:
+            if name not in self._attributes:
+                return None
```
fscherf, line 23: Does this make sense? The reason why this method raises a key error when the given name is not present in ``_attributes`` is that ``.pop()`` is supposed to to return the value of ``name`` and ``.remove()`` does not do this.

lona/html/data_binding/inputs.py
```diff
@@ -38,13 +38,14 @@ def handle_change(self, input_event):
         # The solution for this problem is to don't send patches back to users
         # who issued them.
 
-        self.attributes.__setitem__(
-            'value',
-            input_event.data,
-            issuer=(input_event.connection, input_event.window_id),
-        )
+        with self.lock:
```
fscherf, line 41: Why is this necessary?  ``AttributeDict.__setitem__()`` pulls the lock for us.

lona/html/node.py
```diff
@@ -177,11 +177,8 @@ def ignore(self):
     def ignore(self, value):
         if value:
             self._attributes['data-lona-ignore'] = 'True'
-
```
fscherf, line 180: I want to keep this newline to keep styling consistent with the rest of the code

fscherf (changes requested):
Great changes in general :+1:  I am ok with everything i didn't comment on

lona/html/attribute_dict.py
maratori: What do you think if I change it as following?
1. `attributes.pop(name)` will raise an exception if name is unknown
2. `attributes.pop(name, default)` will not raise an exception, but return `default`
3. `del attributes[name]` will not raise an exception

lona/html/data_binding/inputs.py
maratori: Lock may be dropped if we can drop we can drop `self.value = input_event.data` and use only `self.attributes.__setitem__`.
Otherwise there is a data race between them.

But you are right, that this change should be in a separate PR.

fscherf:
@maratori oh ok. big pr's with smaller commits are also fine

PR #47, opened 2021-09-02
html: implement __contains__ with lock

maratori:
Before this change `__iter__` was used to execute `"disabled" in node.attributes`. And there was a chance to get `RuntimeError: dictionary changed size during iteration` in concurrent execution.
Now the whole operation is protected with lock.

### DIFF
--- a/lona/html/attribute_dict.py
+++ b/lona/html/attribute_dict.py
@@ -111,6 +111,10 @@ class AttributeDict:
         with self._node.lock:
             return self._attributes.__iter__()
 
+    def __contains__(self, name):
+        with self._node.lock:
+            return name in self._attributes
+
     def __bool__(self):
         with self._node.lock:
             return bool(self._attributes)
--- a/lona/html/attribute_list.py
+++ b/lona/html/attribute_list.py
@@ -103,6 +103,10 @@ class AttributeList:
         with self._node.lock:
             return self._attributes.__iter__()
 
+    def __contains__(self, attribute):
+        with self._node.lock:
+            return attribute in self._attributes
+
     # serialisation ###########################################################
     def _reset(self, value):
         if not isinstance(value, list):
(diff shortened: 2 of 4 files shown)


fscherf:
@maratori: Hm. But AttributeDict.__iter__ also pulls a lock. Why does it not work than?

maratori:
@fscherf I've added test that fails without `__contains__`.

Lock in `__iter__` is used only to create dict iterator. But not to iterate over it.
Here is a simple example, to show what's happening:
```py
import threading

lock = threading.Lock()
d = {"a": "aaa", "b": "bbb"}

# call __iter__ in thread A
with lock:
    i = iter(d)

# start searching in thread A
next(i)  # a

# update dict in thread B
with lock:
    d["c"] = "ccc"

# continue searching in thread A
next(i)  # RuntimeError: dictionary changed size during iteration
```

fscherf:
@maratori: I see! I agree with adding ``__contains__`` methods, but i can't get the test you provided to fail without your changes. 
The changes make sense, so you can simply remove the test if you don't want to fix it.

maratori:
> i can't get the test you provided to fail without your changes.

@fscherf That's strange. It fails to me every time.
Did you completely remove `__contains__`?
What python version do you use?

fscherf:
@maratori I checked out your branch, removed ``AttributeDict.__contains__`` (my understanding is that the test only uses this), and ran ``make ci-test`` which runs all supported python versions.

maratori:
Here ci failed. But it fails not every time. https://github.com/maratori/lona/actions/runs/1197181646

Ok, I'll remove the test because it doesn't work as expected.

fscherf:
@maratori I see. Hm seems to be timing sensitive. Ok, the pr looks good to me like this! Thanks!

PR #50, opened 2021-09-02
html: attributes: add default to pop() and make del silent

maratori:
* Added the second optional argument to `AttributeDict.pop`. Now `node.attributes.pop('foo', 'bar')=='bar'` if there is no `foo` in attributes.
* Now `del node.attributes['foo']` do nothing if there is no `foo` in attributes. Before `KeyError` raised.

### DIFF
--- a/lona/html/attribute_dict.py
+++ b/lona/html/attribute_dict.py
@@ -17,8 +17,15 @@ class AttributeDict:
         with self._node.lock:
             return self._attributes.items()
 
-    def pop(self, name):
+    def pop(self, name, *default):
+        if len(default) > 1:
+            raise TypeError(f"pop expected at most 2 arguments, "
+                            f"got {1 + len(default)}")
+
         with self._node.lock:
+            if default and name not in self._attributes:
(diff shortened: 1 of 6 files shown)


fscherf:
@maratori: But python dicts raise a NameError when attempting to delete a non-existing key.

```
>>> a = {'foo': 'bar'}
>>> del a['foo']
>>> a
{}
>>> del a['foo']
  File "/home/fsc/opt/python3/lib/python3.7/site-packages/rlpython/runtimes/python_runtime.py", line 211, in run
    return_value = run_code(code, self.globals, self.locals)
  File "<input>", line 1, in <module>
KeyError: 'foo'
```

I want AttributeDicts to behave like dicts and AttributeLists like lists

fscherf:
@maratori: I revert my earlier comment. AttributeLists and AttributeDicts should behave like the node properties in the browser. ``AttributeList.remove()`` does this already. So this attempt is completely valid

fscherf:
Thanks! :)

PR #51, opened 2021-09-02
html: data binding: always return bool for node.disabled

maratori:
Before that it was possible to return empty string instead of `False`


fscherf:
Ack! Thanks!

PR #52, opened 2021-09-02
Let's increase max line length from 79 to 100 for linter

maratori:


fscherf:
@maratori: I don't want to set the max line length globally to >80, but we can replace ``# NOQA`` with ``# NOQA: E501`` to disable only the line length check for these lines.

maratori:
Ok

PR #55, opened 2021-09-03
html: data binding: fix TextInput.handle_change

fscherf:
TextInput.handle_change gets called at the correct time, but the value of
TextInput.value gets not set before, which is counter-intuitive.

This patch sets TextInput.value in TextInput.handle_input_event() and then
calls TextInput.handle_change(). This patch also adds a test view for this
feature to the test project.

### DIFF
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -7,6 +7,7 @@ DEFAULT_INPUT_DELAY = 300
 class TextInput(Node):
     TAG_NAME = 'input'
     SELF_CLOSING_TAG = True
+    INPUT_ATTRIBUTE_NAME = 'value'
 
     ATTRIBUTES = {
         'type': 'text',
@@ -25,7 +26,7 @@ class TextInput(Node):
         if value is not None:
             self.value = value
 
-    def handle_change(self, input_event):
+    def handle_input_event(self, input_event):
         # Data binding nodes catch their own change events and synchronize
         # their internal value. When setting their value, a HTML patch,
         # containing the set value, gets created that gets distributed to all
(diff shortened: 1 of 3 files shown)


lona/html/data_binding/inputs.py
```diff
@@ -38,6 +38,9 @@ def handle_change(self, input_event):
         # The solution for this problem is to don't send patches back to users
         # who issued them.
 
+        if input_event.name != 'change':
+            return input_event
```
maratori, line 42:
```suggestion
        if input_event.name == 'click':
            # need to take into account self.bubble_up?
            return self.handle_click(input_event)

        if input_event.name != 'change':
            return input_event
```

lona/html/data_binding/inputs.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, value=None, disabled=False, bubble_up=False,
         if value is not None:
             self.value = value
 
-    def handle_change(self, input_event):
+    def handle_input_event(self, input_event):
```
maratori, line 28: Need to update `Select` the same way.

maratori:
@fscherf What do you think about alternative solution?
```diff
Index: lona/view_runtime.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lona/view_runtime.py b/lona/view_runtime.py
--- a/lona/view_runtime.py	(revision 7d765542968e5329f247bce4844774a60437ea64)
+++ b/lona/view_runtime.py	(date 1630657167894)
@@ -692,6 +692,12 @@
 
                 input_event = node.handle_input_event(input_event)
 
+                if input_event.name == 'change':
+                    input_event = node.handle_change(input_event)
+
+                elif input_event.name == 'click':
+                    input_event = node.handle_click(input_event)
+
                 if not input_event:
                     send_html_patches()
 
Index: lona/html/abstract_node.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lona/html/abstract_node.py b/lona/html/abstract_node.py
--- a/lona/html/abstract_node.py	(revision 7d765542968e5329f247bce4844774a60437ea64)
+++ b/lona/html/abstract_node.py	(date 1630675694481)
@@ -92,12 +92,6 @@
         return input_event
 
     def handle_input_event(self, input_event):
-        if input_event.name == 'change':
-            return self.handle_change(input_event)
-
-        elif input_event.name == 'click':
-            return self.handle_click(input_event)
-
         return input_event
 
     # queries #################################################################

```

fscherf:
@maratori I don't like the approach to handle that in the runtime. Also if can't ``node.handle_input_event()`` stop the callback chain after this patch, i read the diff correctly. ``handle_change()`` or ``handle_click()`` are getting called in any case.

I like that in the current design only one method is used to pass events from the runtime to the node. I don't to split that. Also in future there could be more event types besides change and click

maratori:
@fscherf Ok. I like your solution as well.

fscherf:
@maratori I updated the test view so it contains all changeable widgets and applied the changes you suggested

maratori (approved):
Great, thanks!

fscherf:
@maratori Thanks for the review!

PR #56, opened 2021-09-03
html: attributes: optimisation: use set under the hood

### LINKED ISSUE #48: `AttributeList` is a `set` actually, but `list` is used under the hood
`AttributeList` is a `set` actually, but `list` is used under the hood (`AttributeList._attributes`).
This causes performance problems on high number of items.

maratori:
Fixes #48


fscherf:
Great! Thanks! :)

PR #57, opened 2021-09-03
Fscherf/topic/fix custom input events

fscherf:
Custom input events which are not related to a node broke between 1.4.1 and 1.5. This can be reproduced using the test_project under the url ``/frontend/custom-event/``

https://lona-web.org/end-user-documentation/html.html#firing-custom-input-events


maratori (approved):

fscherf:
@maratori thanks for reviewing!

PR #58, opened 2021-09-04
cosmetics: migrate to f-strings

### LINKED ISSUE #54: Minimal python version and f-strings
What is the minimal supported python version?
Can we use f-strings? It would be great if yes.

maratori:
Closes #54

### DIFF
--- a/lona/html/attribute_dict.py
+++ b/lona/html/attribute_dict.py
@@ -73,11 +73,16 @@ class AttributeDict:
         if not isinstance(value, (int, bool, float, str)):
             raise ValueError('unsupported type: {}'.format(type(value)))
 
-        if name in ('id', 'class', 'style'):
+        if name in ('id', 'class'):
             raise RuntimeError(
                 "Node.attributes['{}'] is not supported. "
-                'Use Node.{}{} instead.'.format(
-                    name, name, '_list' if name != 'style' else '')
+                'Use Node.{}_list instead.'.format(name, name)
+            )
+
+        if name == 'style':
+            raise RuntimeError(
+                "Node.attributes['{}'] is not supported. "
+                'Use Node.{} instead.'.format(name, name)
             )
 
         with self._node.lock:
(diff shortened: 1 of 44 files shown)


fscherf:
@maratori Wow thanks! Great changes, only some minor nitpicking from my side :)

lona/html/attribute_dict.py
```diff
@@ -71,13 +71,16 @@ def __getitem__(self, name):
 
     def __setitem__(self, name, value, issuer=None):
         if not isinstance(value, (int, bool, float, str)):
-            raise ValueError('unsupported type: {}'.format(type(value)))
+            raise ValueError(f'unsupported type: {type(value)}')
 
-        if name in ('id', 'class', 'style'):
```
fscherf, line 76: Great change, but this is a functional change. I would like a dedicated commit with commit message for it (just append it to this pr)

lona/html/attribute_dict.py
```diff
@@ -124,15 +127,18 @@ def _reset(self, value):
         if not isinstance(value, dict):
             raise ValueError('unsupported type')
 
-        for k, v in value.items():
+        for name, v in value.items():
             if not isinstance(v, (int, bool, float, str)):
                 raise ValueError('unsupported type')
 
-            if k in ('id', 'class', 'style'):
```
fscherf, line 131: This is also a funtional change

fscherf (changes requested):

fscherf:
@maratori are f-strings a good idea for logging? My understanding of the ``%s`` Syntax in logging is that it does not use python format strings on purpose to have a unformatted string to fall back to if the formatting fails.

maratori:
@fscherf I'm not sure about specific case when formatting fails.
But you are right (new info for me), that %-formatting is better for logging.
Here is a good article: https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/.

lona/html/attribute_dict.py
maratori: Fixed

lona/html/attribute_dict.py
maratori: Fixed

fscherf (approved):
Looks nicer now and everything seems to work like expected

PR #59, opened 2021-09-04
Start using mypy

maratori:
Let's use [mypy](https://github.com/python/mypy). This PR does minimal changes to start:
* ~~add `make mypy`~~
* ~~rename github workflow `test` -> `ci`~~
* run mypy in ci (lint stage)
* run mypy on `make lint`
* add instructions to `How To Contribute`
* add minimal type annotations to pass mypy

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,8 @@ commands =
 deps =
     flake8
     flake8-noqa
+    mypy
 
 commands =
     flake8 --config=flake8.ini lona tests test_project test_script doc
+    mypy -p lona
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ PYTHON_ENV_ROOT=envs
 PYTHON_DEV_ENV=$(PYTHON_ENV_ROOT)/$(PYTHON)-dev
 PYTHON_PACKAGING_ENV=$(PYTHON_ENV_ROOT)/$(PYTHON)-packaging-env
 
-.PHONY: clean doc sdist test ci-test shell freeze
+.PHONY: clean doc sdist test ci-test lint shell freeze
 
 # development environment #####################################################
 $(PYTHON_DEV_ENV)/.created: REQUIREMENTS.dev.txt
@@ -55,7 +55,7 @@ ci-test: dev-env
 # linting #####################################################################
 lint: dev-env
 	. $(PYTHON_DEV_ENV)/bin/activate && \
-	tox -e lint
+	time tox -e lint
 
 # packaging ###################################################################
 sdist: packaging-env
--- a/REQUIREMENTS.dev.txt
+++ b/REQUIREMENTS.dev.txt
@@ -2,3 +2,4 @@
 tox
 flake8
 flake8-noqa
+mypy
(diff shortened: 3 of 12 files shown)


fscherf:
@maratori Great idea! Thanks for bringing this up. I didn't use mypy productive before and i have some questions.

Can we add mypy to the linting stage of the ci, or does this have to be its own stage?

Shouldn't we put mypy config in a mypy.ini instead of appending it to the tox.ini?

maratori:
> Can we add mypy to the linting stage of the ci, or does this have to be its own stage?

Sure, we can.

> Shouldn't we put mypy config in a mypy.ini instead of appending it to the tox.ini?

We can use both of them.
`tox.ini` is one of default config files for [flake8](https://flake8.pycqa.org/en/latest/user/configuration.html).
So I decided to put mypy there as well.

Do you want me to extract config to `mypy.ini`?
And run mypy along with flake8 on lint stage (both ci and make)?

fscherf:
@maratori I would like to have both tools as part of the linting stage and have both of them have their own dedicated config file if possible

tox.ini
```diff
@@ -25,3 +25,16 @@ deps =
...
+deps =
+    mypy
+
+commands =
+    mypy --config-file tox.ini lona
+
+
+[mypy]
+
+[mypy-rlpython.*]
```
fscherf, line 39: What does this do? Why is this necessary for rlpython and not the other dependencies?

maratori: Added comment to `mypy.ini`
```
Avoid error: Skipping analyzing "rlpython": found module but no type hints or library stubs
```

maratori:
@fscherf Fixed

tox.ini
maratori: Here is a run without this config: https://github.com/lona-web-org/lona/runs/3514862356?check_suite_focus=true

fscherf: Ah ok. Do we have to black-list all third party-libraries that produce mypy warnings or can we white-list the lona module instead?

maratori: I'd say that black list is better. It is more strict. But I can change if you want.

fscherf: A black list has to be maintained and the listed modules can change with every test-run. If aiohttp merges a module that throws mypy warnings, our linting stage fails on the next run.

maratori: > If aiohttp merges a module that throws mypy warnings, our linting stage fails on the next run.

As far as I understand it works differently. And that may not happen. However I'm not 100% sure.

Official doc [recommends](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library) to avoid this.
> To suppress all missing import errors for all libraries in your codebase, [...] set the `ignore_missing_imports` config file option to `True` in the global section of your mypy config file.
> [...]
> We recommend using this approach only as a **last resort**: it’s equivalent to adding a # type: ignore to all unresolved imports in your codebase.

fscherf: @maratori That convinces me. Thanks for researching this!

fscherf:
@maratori I am sorry, i merged in the wrong order. Could you rebase this branch so i break nothing? I have nothing more to complain on, so i will merge after.

maratori:
@fscherf 👌

fscherf:
@maratori Great changes! Thanks for adding this! :)

PR #60, opened 2021-09-04
cosmetics: use %r for logging instead of repr()

maratori:
It also may increase performance a bit because of lazy formatting. `repr` may even not be called depending on selected log level.


fscherf:
@maratori Great! Thanks! :)

PR #61, opened 2021-09-05
Fscherf/topic/fix boolean properties

### LINKED ISSUE #49: Boolean attributes doesn't work
Some "boolean" attributes doesn't work as expected. Below is an example.
```py
from lona import LonaApp
from lona.html import HTML, Select, Button
from lona.view import LonaView

app = LonaApp(__file__)

@app.route("/")
class AttributesWithoutValue(LonaView):
    def handle_request(self, request):
        return HTML(
            Select(values=[(1, "aaa"), (2, "bbb")], multiple=False),  # actually it is multiple in browser
            Button("xxx", disabled=False),  # actually it is disabled in browser
        )

app.run(port=8080)
```

@fscherf What do you think if we will handle such special attributes (`disabled`, `checked`, `multiple`, ...) in `AttributesDict` class?
For example if I execute `node.attributes["disabled"] = False` it will remove `disabled` attribute from html.

fscherf:
This fixes #49

### DIFF
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -15,7 +15,7 @@ class TextInput(Node):
         'type': 'text',
     }
 
-    def __init__(self, value=None, disabled=False, bubble_up=False,
+    def __init__(self, value=None, disabled=None, bubble_up=False,
                  input_delay=DEFAULT_INPUT_DELAY, **kwargs):
 
         super().__init__(**kwargs)
@@ -23,10 +23,12 @@ class TextInput(Node):
         self.events.add(CHANGE(input_delay))
 
         self.bubble_up = bubble_up
-        self.disabled = disabled
 
         if value is not None:
             self.value = value
+
+        if disabled is not None:
+            self.disabled = disabled
 
     def handle_input_event(self, input_event):
         # Data binding nodes catch their own change events and synchronize
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -78,6 +78,9 @@ class TextInput(Node):
 
     @disabled.setter
     def disabled(self, new_value):
+        if not isinstance(new_value, bool):
+            raise TypeError('disabled is a boolean property')
+
         if new_value:
             self.attributes['disabled'] = True
 
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -82,7 +82,7 @@ class TextInput(Node):
             raise TypeError('disabled is a boolean property')
 
         if new_value:
-            self.attributes['disabled'] = True
+            self.attributes['disabled'] = ''
 
         else:
             del self.attributes['disabled']
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -15,8 +15,8 @@ class TextInput(Node):
         'type': 'text',
     }
 
-    def __init__(self, value=None, disabled=None, bubble_up=False,
-                 input_delay=DEFAULT_INPUT_DELAY, **kwargs):
+    def __init__(self, value=None, disabled=None, readonly=None,
+                 bubble_up=False, input_delay=DEFAULT_INPUT_DELAY, **kwargs):
 
         super().__init__(**kwargs)
 
@@ -29,6 +29,9 @@ class TextInput(Node):
 
         if disabled is not None:
             self.disabled = disabled
+
+        if readonly is not None:
+            self.readonly = readonly
 
     def handle_input_event(self, input_event):
         # Data binding nodes catch their own change events and synchronize
@@ -86,6 +89,22 @@ class TextInput(Node):
 
         else:
             del self.attributes['disabled']
+
+    # readonly
+    @property
+    def readonly(self):
+        return 'readonly' in self.attributes
+
+    @readonly.setter
+    def readonly(self, new_value):
+        if not isinstance(new_value, bool):
+            raise TypeError('readonly is a boolean property')
+
+        if new_value:
+            self.attributes['readonly'] = ''
+
+        else:
+            del self.attributes['readonly']
 
 
 class TextArea(TextInput):
--- a/lona/html/node.py
+++ b/lona/html/node.py
@@ -181,6 +181,9 @@ class Node(AbstractNode):
 
     @ignore.setter
     def ignore(self, value):
+        if not isinstance(value, bool):
+            raise TypeError('ignore is a boolean property')
+
         if value:
             self._attributes['data-lona-ignore'] = 'True'
 
--- a/lona/html/node.py
+++ b/lona/html/node.py
@@ -10,6 +10,23 @@ from lona.html.node_list import NodeList
 from lona.protocol import NODE_TYPE
 
 
+def parse_style_string(style_string: str) -> Dict[str, str]:
+    values = {}
+
+    for css_rule in style_string.split(';'):
+        if not css_rule.strip():
+            continue
+
+        if ':' not in css_rule:
+            raise ValueError(f'Invalid style string: {style_string}')
+
+        name, value = css_rule.split(':', 1)
+
+        values[name.strip()] = value.strip()
+
+    return values
+
+
 class Node(AbstractNode):
     TAG_NAME = 'html'
     SELF_CLOSING_TAG = False
@@ -87,6 +104,9 @@ class Node(AbstractNode):
             elif name == 'style':
                 if not isinstance(value, (dict, str)):
                     raise ValueError('style has to be dict or string')
+
+                if isinstance(value, str):
+                    value = parse_style_string(value)
 
                 self._style.update(value)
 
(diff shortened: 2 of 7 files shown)


fscherf:
@maratori That should be the last known bug in the current release. When this got through i will release 1.5.2

lona/html/data_binding/inputs.py
```diff
@@ -74,12 +79,31 @@ def disabled(self):
 
     @disabled.setter
     def disabled(self, new_value):
+        if not isinstance(new_value, bool):
+            raise ValueError('disabled is a boolean property')
```
maratori, line 83: This doesn't allow to use string html
```py
HTML('<input type="checkbox" disabled/>')
```

maratori (changes requested):

lona/html/data_binding/inputs.py
maratori: I haven't created an issue yet. But seems it's good time to say that.

I think it's better to use `node.attributes` when parsing html string instead of passing `kwargs` to constructor here:
https://github.com/lona-web-org/lona/blob/049201b78772ede3a83590ae54958872fb1c4b49/lona/html/parsing.py#L129

maratori: For example following `HTML('<div nodes="aaa">bbb</div>')` produces `Div("aaa", "bbb")`, but I expect to get `Div("bbb", attributes={"nodes": "aaa"})`.
I expect that string will be treated as html. Of course it's ok if lona will add some attributes like `data-lona-node-id` and `data-lona-events`.

fscherf: Oh! You are right! I liked that the high level properties get used to set boolean attributes, but you would have to make many exceptions to make this work. Ok we need to use ``Node.attributes`` instead of ``Node.__init__`` to set data.

Should we open an dedicated issue for this or should i append changes for that here?

maratori: I don't mind to see changes in this PR.

fscherf: @maratori fixed! Thanks for bringing this up

maratori: Here will be `text` always because `attributes` now contains only `tag_name` and `self_closing_tag`:
https://github.com/lona-web-org/lona/blob/808c42c96a5faaa6ef11ee2ce6a0c4c4df2545a1/lona/html/parsing.py#L94-L95

lona/html/parsing.py
```diff
@@ -111,23 +111,22 @@ def get_node_class(self, tag_name, attributes):
...
 
         # setup node
         node_class = self.get_node_class(tag, node_kwargs)
         node = node_class(**node_kwargs)
 
+        for name, value in attrs:
+            if value is None:
+                value = 'true'
+
+            node.attributes[name] = value
```
maratori, line 128: This fails for following attributes:
```py
HTML('<div id="abc">')
HTML('<div class="abc">')
HTML('<div style="color: red;">')
```

fscherf: @maratori Ack -.- we need tests for all these cases

fscherf:
@maratori I think i got everything sorted out now and added a simple test

fscherf:
@maratori I think we should remove the ```Node.style = 'color: black'``` Syntax and make it consistent with the rest of the API again

lona/html/parsing.py
```diff
@@ -107,24 +108,46 @@ def get_node_class(self, tag_name, attributes):
...
         # setup node
         node_class = self.get_node_class(tag, node_kwargs)
         node = node_class(**node_kwargs)
 
+        for name, value in attrs:
+            if value is None:
+                value = ''
+
+            if name == 'id':
+                node.id_list.extend(self.parse_attribute_list(value))
```
maratori, line 140: nit: Seems this should work as well
```suggestion
                node.id_list.extend(value.split())
```

maratori:
> @maratori I think we should remove the `Node.style = 'color: black'` Syntax and make it consistent with the rest of the API again

@fscherf Agree. One way of doing things is more simple.

lona/html/data_binding/inputs.py
```diff
@@ -76,12 +81,31 @@ def disabled(self):
...
 
+    # readonly
+    @property
+    def readonly(self):
+        return 'readonly' in self.attributes
+
+    @readonly.setter
+    def readonly(self, new_value):
+        if not isinstance(new_value, bool):
+            raise ValueError('readonly is a boolean property')
```
maratori, line 101: nit:
```suggestion
            raise TypeError('readonly is a boolean property')
```

lona/html/node.py
```diff
@@ -181,6 +181,9 @@ def ignore(self):
 
     @ignore.setter
     def ignore(self, value):
+        if not isinstance(value, bool):
+            raise ValueError('ignore is a boolean property')
+
         if value:
             self._attributes['data-lona-ignore'] = 'True'
```
maratori, line 188: nit:
```suggestion
            self._attributes['data-lona-ignore'] = ''
```

tests/test_0001_html.py
```diff
@@ -258,3 +258,86 @@ def test_attribute_list():
     d.id_list.toggle('foo')
 
     assert d.id_list == ['foo', 'bar']
+
+
+@pytest.mark.dependency()
+def test_html_strings():
```
maratori, line 264: nit: I think it's better to split this to several small tests (you already did it using `######`).
Also it's better to move those tests to new `test_0002_parsing_html.py`.

tests/test_0001_html.py
```diff
@@ -258,3 +258,86 @@ def test_attribute_list():
     d.id_list.toggle('foo')
 
     assert d.id_list == ['foo', 'bar']
+
+
+@pytest.mark.dependency()
```
maratori, line 263: nit: Do we need add this line to each test?

maratori (approved):
@fscherf Great work! I only have few nitpicking comments.

lona/html/parsing.py
fscherf: @maratori Ouh! I didn't know that ``str.split()`` can remove empty strings

tests/test_0001_html.py
fscherf: @maratori I like that parsing is only one test, because if you break it, only one test fails instead of multiple. In big test-suites its often the case that you change one thing and the whole suite goes red. You can't  see hot-spots that easy. I tried to put this test not into its own file because pytest-dependency is not very clever when searching for tests in other modules.

tests/test_0001_html.py
fscherf: @maratori Currently yes. I am not very happy with how this works right now, but the idea was that you can have low-level tests (pretty much all tests at the moment are low-level tests) and high-level tests that involve a browser etc, and high-level tests don't run if the low-level tests fail.

fscherf:
@maratori I added all your requested changes and moved style string parsing to ``lona.html.Node``, so it's consistent with the rest of the api.

lona/html/node.py
```diff
@@ -10,6 +10,23 @@
 from lona.protocol import NODE_TYPE
 
 
+def parse_style_string(style_string: str) -> Dict[str, str]:
```
maratori, line 13: Hm... Now I'm not so sure about this design decision.

Ok we have consistency between `Node.id_list`, `Node.class_list` and `Node.style`.
But there is no consistency between those properties and `Node.__init__`.

So this works well:
```py
Div(id='foo bar', _class='foo bar', style='color: red')
```
But this doesn't work:
```py
Div().id_list = 'foo bar'
Div().class_list = 'foo bar'
Div().style = 'color: red'
```

@fscherf What do you think?

From the other side `id` from constructor becomes `id_list`. This stresses out that it should be a list. So maybe it's ok.

maratori (approved):

lona/html/node.py
fscherf: @maratori 

> From the other side id from constructor becomes id_list. This stresses out that it should be a list. So maybe it's ok.

That was what i thought too. I experimented with the corresponding Javascript API and the Lona API should mimic ``element.idList``, ``element.classList`` and ``element.style``. I think this implementation is closest to the Javascript equivalent.

lona/html/parsing.py
```diff
@@ -108,23 +108,33 @@ def get_node_class(self, tag_name, attributes):
...
+            if key in node_attributes:
+                node_kwargs[key] = node_attributes.pop(key)
+
         node_class = self.get_node_class(tag, node_kwargs)
         node = node_class(**node_kwargs)
 
+        # set attributes
+        for name, value in node_attributes.items():
+            node.attributes[name] = value
+
```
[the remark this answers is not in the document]
fscherf: @maratori oh good point!

maratori (approved):

maratori:
@fscherf great changes! Huge improvement, thanks.

fscherf:
@maratori Thanks for reviewing! :) I will relase 1.5.2 so we have a stable release again.

PR #62, opened 2021-09-05
lint: use flake8-noqa plugin

maratori:
See https://github.com/plinss/flake8-noqa/


fscherf:
@maratori Great!  I have nothing to complain about :) Thanks!

PR #63, opened 2021-09-05
html: Submit, Reset: remove lona click events

fscherf:
Since we restructured click and change events, Submit and Reset catch their own click events, which breaks traditional froms
(test project url: /view-types/form-view/)

### DIFF
--- a/lona/html/nodes.py
+++ b/lona/html/nodes.py
@@ -161,7 +161,6 @@ class Button(Node):
 class Submit(Node):
     TAG_NAME = 'input'
     SELF_CLOSING_TAG = True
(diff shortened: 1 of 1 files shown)


maratori:
@fscherf I'd say we need to remove `Reset` button at all because:
1. It doesn't work as expected in lona
2. Currently [you should usually avoid including reset buttons in your forms](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/reset)

maratori (approved):

fscherf:
@maratori I agree. I added a patch that removes the reset node

maratori (approved):
👍

PR #64, opened 2021-09-05
lint: use flake8-quotes plugin

maratori:
See https://github.com/zheller/flake8-quotes/


fscherf:
@maratori: Great idea! :)

PR #65, opened 2021-09-05
lint: use flake8-comprehensions plugin

maratori:
See https://github.com/adamchainz/flake8-comprehensions


fscherf:
@maratori ack! Thanks :)

PR #66, opened 2021-09-05
lint: use flake8-pytest-style plugin

maratori:
See https://github.com/m-burst/flake8-pytest-style


fscherf:
@maratori Ouh! That looks great! :)

PR #67, opened 2021-09-05
lint: use flake8-logging-format plugin

maratori:
See https://github.com/globality-corp/flake8-logging-format


fscherf:
@maratori: great! Thanks :)

PR #68, opened 2021-09-05
lint: use flake8-bugbear plugin

maratori:
See https://github.com/PyCQA/flake8-bugbear

### DIFF
--- a/REQUIREMENTS.dev.txt
+++ b/REQUIREMENTS.dev.txt
@@ -4,6 +4,7 @@ pytest
 pytest-cov
 pytest-dependency
 flake8
+flake8-bugbear
 flake8-comprehensions
 flake8-logging-format
 flake8-noqa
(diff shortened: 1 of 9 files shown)


fscherf:
@maratori this list of plugins gets pretty long. I have no experience with most of them. Are you willing to maintain this list? Or maybe linting as whole?

maratori:
@fscherf sure, I can maintain them.
I'm not an expert in python linters either (yet :)
However I have a lot of experience with linters in golang.

I add only linters with minimal impact. Most probably we will not have any problems with them.

fscherf:
@maratori cool :) Then i am ok with all the plugins :)

fscherf:
@maratori: Why do you add variables with leading underscore in some hunks, and remove it in others?

maratori:
@fscherf Good question. I no answer :)
I reverted unnecessary changes.
Underscore in variable name stresses that it is not used.

fscherf:
@maratori Looks much cleaner now :) Thanks!

PR #69, opened 2021-09-05
dev: add type annotations to app

maratori:


lona/app.py
```diff
@@ -72,40 +79,73 @@ def resolve_path(self, path):
...
 
-    def middleware(self, middleware_class=None):
-        def decorator(middleware_class):
+    @overload
+    def middleware(self) -> Callable[[Type], None]: ...
+
+    @overload
+    def middleware(self, middleware_class: Type) -> None: ...
+
+    def middleware(self, middleware_class: Optional[Type] = None) -> \
```
fscherf, line 111: In cases like this i would rather add ``# NOQA: $ERROR_CODE``

lona/app.py
```diff
@@ -157,7 +207,7 @@ def add_static_file(self, name, string='', path=''):
         )
 
     # server ##################################################################
-    def run(self, **args):
+    def run(self, **args: Any) -> None:  # TODO: change to explicit arguments
```
fscherf, line 210: Does this mean your pr is not ready, or do you want to add this todo string?

lona/app.py
maratori: Does your comment relates to overloading or to multiline declaration?

fscherf: @maratori Multiline declaration. I don't like the ``\`` syntax

maratori: Ok. I'll rebase this after #75 is merged.

maratori: I've decided to not wait for #75.

lona/app.py
maratori: Removed comment

lona/app.py
```diff
@@ -72,40 +83,71 @@ def resolve_path(self, path):
...
+            decorator(middleware_class)
+            return None
 
         else:
             return decorator
 
-    def frontend_view(self, view_class=None):
-        def decorator(view_class):
+    @overload
+    def frontend_view(self) -> Callable[[Type[LonaView]], None]: ...
```
fscherf, line 127: @maratori I don't like the one-line syntax. Could you change it to two separate lines?

setup.py
```diff
@@ -18,6 +18,8 @@
         'aiohttp>=3,<4',
         'jinja2',
         'rlpython',
+        'typing',
+        'typing-extensions',
```
fscherf, line 22: @maratori This is not part of the python standard lib?

fscherf (changes requested):

lona/app.py
maratori: Sure, fixed.

setup.py
maratori: You are right, `typing` is stdlib (I removed it).
But `typing-extensions` isn't.
```
$ docker run --rm python python -c 'import typing_extensions'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'typing_extensions'
```

fscherf: @maratori the documentation of typing-extensions states 

>Typing Extensions – Backported and Experimental Type Hints for Python

We don't need backports, or do we? And should we use experimental features?

maratori: We don't use experimental features, only one backported type `Literal`.
It was introduced in [Python 3.8](https://docs.python.org/3/library/typing.html#typing.Literal).
So we have to use `typing-extensions`, see https://github.com/lona-web-org/lona/pull/69/files#diff-95f7c4beb7cedde530d15bdf9b7c86d405985af7b5fd18ad067728b756f80e7bR11.

fscherf: Ah ok! Thanks for the explanation

fscherf:
@maratori thanks! :)

PR #70, opened 2021-09-06
html: accept handle_change and handle_click in node constructor

maratori:


fscherf:
@maratori Ah! Like in your original proposal. I will merge this into 1.5.2 as well, because it's a very small change

PR #72, opened 2021-09-07
client: rendering: fix handling of boolean attributes

fscherf:


lona/client/dom-renderer.js
```diff
@@ -5,7 +5,9 @@ Lona.LonaDomRenderer = function(lona_context, lona_window) {
     // html rendering ---------------------------------------------------------
     this._render_node = function(node_spec) {
         var _this = this;
-        var property_names = ['value', 'checked'];
+
+        var property_names = ['value', 'disabled', 'checked', 'selected',
+                              'readonly'];
```
maratori, line 10: Why do we need `disabled` and `readonly` in this list?

lona/client/dom-updater.js
```diff
@@ -439,15 +441,20 @@ Lona.LonaDomUpdater = function(lona_context, lona_window) {
...
+                    node[name] = value;
 
                 // attributes
                 } else {
-                    node.setAttribute(data[0], data[1]);
-                }
+                    node.setAttribute(name, data[1]);
+                };
 
             // RESET
```
maratori, line 459: Don't we need to do the same check on reset?

lona/client/dom-updater.js
```diff
@@ -467,7 +474,16 @@ Lona.LonaDomUpdater = function(lona_context, lona_window) {
...
+                // properties
+                if(property_names.indexOf(name) > -1) {
+                    node[name] = false;
+
+                // attributes
+                } else {
+                    node.removeAttribute(name);
+                };
 
             // CLEAR
```
maratori, line 488: Don't we need to do the same check on clear?

lona/client/dom-renderer.js
fscherf: The browser API is very strange in this regard. ``button.disabled`` and ``button.setAttribute('disabled', '')``  both work, for ``checkbox`` and ``checked`` only the first syntax works. I am not sure which syntax makes more here. Maybe we just continue to handle ``value`` and ``checked`` separately.

lona/client/dom-updater.js
fscherf: Ack! I will fix that.

lona/client/dom-updater.js
fscherf: Ack! I will fix that.

lona/client/dom-renderer.js
maratori: My imagination says that there are two types of properties: "static" and "dynamic" (naming is mine :).

"Static" means that data is stored in html attributes (ex. `disabled`). In another words there is a two way coupling between html attributes and js properties. They are always in sync.

"Dynamic" means that there are two independent data: html attribute and js property (ex. `value`). There is only one type of connection between them: html attribute defines initial value for js property. That's how reset button works for example.

fscherf: Makes sense. But who decided that ``disabled`` is "static" but ``checked`` not!?
Ok, so we keep this list to "dynamic" properties (``value`` and ``checked``) ?

maratori: I'd say that all properties user can change is "dynamic": `value`, `checked`, `selected`. Others are "static".

fscherf: Ohhhhhh!!!! \*facepalm\*
Ok! Then this list should contain ``value``, ``checked`` and ``selected`` right?

maratori: In my understanding yes.
But I'm not an expert in HTML and JS :)

fscherf: me neither, but it make sense to me :) Also it works!
I changed it that way.

lona/client/dom-updater.js
```diff
@@ -458,16 +463,37 @@ Lona.LonaDomUpdater = function(lona_context, lona_window) {
...
                 };
 
             // REMOVE
             } else if(operation == protocol.OPERATION.REMOVE) {
-                node.removeAttribute(data[0]);
+                var name = data[0];
+
+                // properties
+                if(property_names.indexOf(name) > -1) {
+                    node[name] = false;
```
maratori, line 491:
```suggestion
                    if(name == 'value') {
                        node[name] = "";
                    } else {
                        node[name] = false;
                    }
```

fscherf: fixed

maratori: Have you pushed changes?

fscherf: @maratori yep! But to the wrong remote :)

maratori (approved):
@fscherf great, thanks!

fscherf:
@maratori thanks for reviewing!

PR #75, opened 2021-09-07
tests: setup browser testing using playwright

fscherf:


tox.ini
```diff
@@ -14,8 +14,11 @@ deps =
     pytest
     pytest-cov
     pytest-dependency
+    pytest-aiohttp
```
maratori, line 17: nit: Don't we need to add them to `REQUIREMENTS.dev.txt` as well?

tests/test_click_events.py
```diff
@@ -0,0 +1,31 @@
...
+    from lona import LonaView
+
+    @app.route('/')
+    class MyLonaView(LonaView):
+        def handle_request(self, request):
+            self.show(Button('click me!'))
+            self.await_click()
+            self.show('SUCCESS')
+
+    app.settings.SESSIONS = False
```
maratori, line 12: nit: Why is it necessary?

tests/test_click_events.py
```diff
@@ -0,0 +1,31 @@
...
+    async with async_playwright() as p:
+        browser = await p.chromium.launch()
+        page = await browser.new_page()
+        await page.goto(context.get_url())
+
+        assert await page.inner_text('#lona') != 'SUCCESS'
+
+        await page.click('button')
+
+        assert await page.inner_text('#lona') == 'SUCCESS'
```
maratori, line 29: It's better to use methods that can wait for desired state.
```suggestion
        await page.wait_for_selector('#lona:has-text("SUCCESS")')
```

For example if I add `self.sleep(1)` before `self.show('SUCCESS')` in view handler, than suggested variant will pass, but initial one will fail.

.github/workflows/test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9']
```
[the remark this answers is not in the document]
fscherf: Running only half of the test-suite has also a potential for false-negative test results. We could drop python3.6 support all together and have Lona require python3.7+

lona/pytest.py
```diff
@@ -0,0 +1,35 @@
+import pytest
```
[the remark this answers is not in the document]
fscherf: It's no real dependency, because the module is standalone and gets ignored by Lona itself. It's only used in test-environments.

setup.py
```diff
@@ -22,4 +22,9 @@
     scripts=[
         'bin/lona',
     ],
+    entry_points={
+        'pytest11': [
+            'lona = lona.pytest',
```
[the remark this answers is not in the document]
fscherf: Both. I want to use this plugin for tests in Lona and in Lona based projects

lona/pytest.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def get_url(self):
+        return f'http://{self.client.host}:{self.client.port}'
+
+
+@pytest.fixture()
+async def lona_app_context(aiohttp_client):
+    async def setup_lona_app_context(setup_app):
+        # setup lona app
+        lona_app = LonaApp(__file__)
```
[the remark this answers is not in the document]
fscherf: Good idea! And it should be resetable by the test.

tox.ini
```diff
@@ -14,8 +14,11 @@ deps =
...
+    playwright
 
 commands =
+    python -m playwright install
```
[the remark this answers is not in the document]
fscherf: What does ``install-deps`` do?

tox.ini
fscherf: No. ``REQUIREMENTS.dev.txt`` needs only ``tox`` and Lona itself, to make ``make test`` work. The make target uses ``REQUIREMENTS.dev.txt`` to install tox, which manages its own dependencies.

Btw: ``REQUIREMENTS.dev.txt`` seems to have a lot of obsolete dependencies. I didn't notice that while merging.

tests/test_click_events.py
fscherf: The Lona session middleware checks for a specific cookie to authenticate users. If the cookie is not set, the middleware sets it and redirects the browser, so the cookie gets active. The playwright browser seems not to follow this redirect. Maybe this has to be enabled, or the middleware has to be fixed.

tests/test_click_events.py
```diff
@@ -0,0 +1,31 @@
...
+
+async def test_rendering(lona_app_context):
+    from playwright.async_api import async_playwright
+
+    context = await lona_app_context(setup_app)
+
+    async with async_playwright() as p:
+        browser = await p.chromium.launch()
+        page = await browser.new_page()
+        await page.goto(context.get_url())
```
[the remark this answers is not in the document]
fscherf: Ah ok. If we can't use ``pytest-playwright`` we should add our own fixture that does this kind of things.

tests/test_click_events.py
fscherf: Ouh! :) Good to know! To be honest: This test was a quick-and-dirty proof-of-concept, and i was very happy that it does something.

tests/test_click_events.py
maratori: Test passes on my machine without this like.
According to [docs](https://playwright.dev/python/docs/api/class-page#page-goto), redirection should work well.
> Returns the main resource response. In case of multiple redirects, the navigation will resolve with the response of the last redirect.

tox.ini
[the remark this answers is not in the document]
maratori: I don't know, actually. But comment says
> Install operating system dependencies

Seems it installs something for Windows and Linux
https://github.com/microsoft/playwright/blob/77b3b0965af1cf28670f9509fa265c2cea78e315/src/utils/registry.ts#L525-L528

tests/test_click_events.py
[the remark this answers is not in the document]
fscherf: I added it to the test

tests/test_click_events.py
fscherf: fixed

tests/test_click_events.py
fscherf: You are right, on mine too now. Then i did something wrong yesterday. I removed the line.

tox.ini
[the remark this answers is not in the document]
fscherf: The docs say 
> Ubuntu 18.04 and Ubuntu 20.04 system dependencies can get installed automatically. This is useful for CI environments.
Whatever this means :D i added it.

tox.ini
maratori: Oh, I see. I can share, why I added them in my PRs.
I created venv using `make dev-env` and PyCharm couldn't resolve some imports.

fscherf: Ah ok! Hm. Should we leave them there?

maratori: I don't know. I didn't work with tox. In all my projects I used pipenv.

mypy.ini
```diff
@@ -11,3 +11,7 @@ warn_return_any = True
 [mypy-rlpython.*]
 # Avoid error: Skipping analyzing "rlpython": found module but no type hints or library stubs
 ignore_missing_imports = True
+
+[mypy-pytest.*]
+# Avoid error: Skipping analyzing "rlpython": found module but no type hints or library stubs
+ignore_missing_imports = True
```
maratori, line 17: Let's remove this config and istead add `pytest` in deps of `[testenv:lint]`

tox.ini
fscherf: @maratori Do you run pytest directly when using the test-suite?

lona/pytest.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, client, app):
+        self.client = client
+        self.app = app
+
+    def get_url(self):
+        return f'http://{self.client.host}:{self.client.port}'
+
+
+@pytest.fixture()
+async def lona_app_context(aiohttp_client):
```
maratori, line 16: I'm not an expert in async python. Am I right that here it is not necessary?
```suggestion
def lona_app_context(aiohttp_client):
```

tests/test_click_events.py
```diff
@@ -0,0 +1,26 @@
...
+
+async def test_rendering(lona_app_context):
+    from playwright.async_api import async_playwright
+
+    context = await lona_app_context(setup_app)
+
+    async with async_playwright() as p:
+        browser = await p.chromium.launch()
+        page = await browser.new_page()
+        await browser.new_context()
```
maratori, line 21:
```suggestion
        browser_context = await browser.new_context()
        page = await browser_context.new_page()
```

tox.ini
maratori: In lona I use `make test`.

fscherf: Ok. Hm then i would say we remove it from ``REQUIREMENTS.dev.txt``

lona/pytest.py
fscherf: You are right. I made it a sync function

tests/test_click_events.py
fscherf: Ah! That makes more sense. Fixed

setup.py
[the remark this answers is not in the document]
maratori: > Has a separate package an advantage that i miss?

Good question. Right now I'd say no (maybe in the future I found something :).
I thought pytest becomes a dependency of lona. But if it's not the case, then I'm ok with that.

mypy.ini
maratori: Wow. I've found that the same works with `rlpython`. We can remove its config as well.

lona/pytest.py
```diff
@@ -0,0 +1,35 @@
...
+from lona import LonaApp
+
+
+class LonaContext:
+    def __init__(self, client, app):
+        self.client = client
+        self.app = app
+
+    def get_url(self):
+        return f'http://{self.client.host}:{self.client.port}'
```
maratori, line 12: nit:
```suggestion
        return str(self.client.make_url(""))
```

lona/pytest.py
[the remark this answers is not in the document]
fscherf: Done

tox.ini
fscherf: Done

lona/pytest.py
fscherf: Good point! Done

lona/pytest.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_url(self):
+        return str(self.client.make_url(''))
+
+
+@pytest.fixture()
+def lona_app_context(request, aiohttp_client):
+    async def setup_lona_app_context(setup_app):
+        # setup lona app
+        project_root = os.path.dirname(request.node.module.__file__)
+        lona_app = LonaApp(project_root)
```
maratori, line 22:
```suggestion
        lona_app = LonaApp(request.fspath)
```

maratori (approved):
Great step towards testing!

tests/test_click_events.py
maratori: I know that it's a POC. But I had a lot of troubles with that, so I had to say that. Sorry :)

lona/pytest.py
fscherf: Ouh! Even better! I changed it that way, and made ``project_root`` configurable.

fscherf:
@maratori I am pretty happy with this infrastructure. If we choose this approach, i will add ``lona_project_context`` as well to make testing against existing Lona projects possible.

lona/pytest.py
maratori: Is `project_root` a good name?
As far as I understand it has to be a path to a file. But reading this name I think it should be a path to directory.

.github/workflows/test.yml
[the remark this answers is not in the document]
fscherf: @maratori what's your opinion on dropping 3.6 support?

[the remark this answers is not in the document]
maratori: I don't mind.
But for this question I'm not the best person to ask :)

fscherf:
@maratori i added a fixture to test Lona projects as well

lona/pytest.py
```diff
@@ -0,0 +1,83 @@
...
+        def setup_aiohttp_app(loop):
+            lona_app.setup_server(loop=loop)
+
+            return lona_app.aiohttp_app
+
+        client = await aiohttp_client(setup_aiohttp_app)
+
+        return LonaContext(
+            client=client,
+            app=lona_app,
```
maratori, line 37: Let's add `server` for consistency.

lona/pytest.py
```diff
@@ -0,0 +1,83 @@
...
+def lona_project_context(request, aiohttp_client):
+    async def setup_lona_project_context(
+        project_root='',
+        settings=None,
+        settings_pre_overrides=None,
+        settings_post_overrides=None,
+    ):
+
+        # setup aiohttp app
+        variables = {
```
maratori, line 53: nit: let's use `nonlocal` instead of dict.

lona/pytest.py
```diff
@@ -0,0 +1,83 @@
...
+
+            variables['server'] = server
+
+            return aiohttp_app
+
+        client = await aiohttp_client(setup_aiohttp_app)
+
+        return LonaContext(
+            client=client,
+            server=variables['server'],
```
maratori, line 80: Am I right that `LonaApp` is not created in this case?

fscherf: @maratori That's right. The app class is just a wrapper around the server to make scripts work. When using a whole project, there is no need to have an app around it.

lona/pytest.py
fscherf: I agree. Fixed

lona/pytest.py
fscherf: Good Idea! Fixed.

lona/pytest.py
fscherf: I initialized it with ``None`` for consistency.

.github/workflows/test.yml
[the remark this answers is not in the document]
fscherf: I think we just drop 3.6 support entirely from Lona. It's life span ends in 3 months anyway.

maratori (approved):
That's great!

maratori:
@fscherf Are you going to add more tests in this PR?

fscherf:
@maratori: @sobolevn didn't post his proposal yet. I want to start adding more tests when we agreed on a test framework. These two tests are just a proof of concept for now.

PR #76, opened 2021-09-09
tests: remove unnecessary mark.dependency

maratori:
See https://pytest-dependency.readthedocs.io/en/stable/configuration.html#configuration-file-options


fscherf:
@maratori Great! Thanks for researching this! :)

PR #77, opened 2021-09-09
doc: setup demos

fscherf:
This PR adds some demos to lona-web.org. To test them, go to ``doc`` and run ``make server``


maratori (approved):
👍

fscherf:
@maratori thanks for reviewing:)

themutt:
I've tested them all they work perfectly! I also really like the use of GIFs. I have an idea for an additional demo to showcase the lona features. Perhaps it's better to submit as a separate PR?

fscherf:
@themutt that would be great! :) thanks in advance! You can provide just the code, i can create a gif for it :)

themutt:
> @themutt that would be great! :) thanks in advance! You can provide just the code, i can create a gif for it :)

https://gist.github.com/themutt/dbd0e45cdb1f00d01a19af9795ca1e09

Here are my two snippets I created for myself from your chat and server state examples. The snippets are mainly focused on synchronizing different views (for example by using `LonaView.show()` without arguments that is a little bit unclear to me) with the same server data. I don't know if there are different ways to accomplish the same goal but feel free to include them if you think they can be useful

themutt (approved):

fscherf:
@themutt Great idea! I didn't know about github discussions. I created one :)

PR #80, opened 2021-09-11
Typo fixes

fscherf:


lona/protocol.py
```diff
@@ -1,6 +1,6 @@
+from typing import Tuple, Optional, Dict, Union
```
maratori, line 1: nit: What was wrong with alphabetical the order?

maratori (approved):

lona/protocol.py
fscherf: All other modules use reverse christmas tree ordered imports (long lines before short lines)

maratori: Wow. Didn't know even about such kind of sorting.
Let's use [linter](https://github.com/gforcada/flake8-isort) to keep it consistent (in separate PR).

fscherf: I learned that when i was studying, in a seminar about C. I thought everyone does this, but after some googling i realize its not that common :D (seems to be a linux style guide thing). I think isort is good idea to get it consistent.

maratori: I've created #105

PR #83, opened 2021-09-12
html: parsing: fix input type definition

maratori:
Before wrong dict was used to get node class. That caused incorrect parsing of `input` HTML element. `TextInput` was used always despite `type` attribute.

### DIFF
--- a/tests/test_0001_html.py
+++ b/tests/test_0001_html.py
@@ -257,7 +257,15 @@ def test_attribute_list():
 
 
 def test_html_strings():
-    from lona.html import HTML, Button
+    from lona.html import (
+        TextInput,
+        CheckBox,
+        TextArea,
+        Button,
+        Select,
+        HTML,
+        Node,
+    )
 
     # empty node ##############################################################
     node = HTML('<div></div>')[0]
@@ -326,7 +334,7 @@ def test_html_strings():
     # high level nodes ########################################################
     node = HTML('<button>Click me</button>')[0]
 
-    assert isinstance(node, Button)
+    assert type(node) is Button
 
     # boolean properties ######################################################
     node = HTML('<button disabled>Click me</button>')[0]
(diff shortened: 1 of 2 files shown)


tests/test_0001_html.py
```diff
@@ -257,7 +257,8 @@ def test_attribute_list():
 
 
 def test_html_strings():
-    from lona.html import HTML, Button
+    from lona.html import HTML, Button, TextInput, TextArea, CheckBox, \
```
fscherf, line 260: please use parentheses in cases like this to keep it consistent with the rest of the code base

tests/test_0001_html.py
```diff
@@ -336,3 +337,46 @@ def test_html_strings():
...
+    node = HTML('<input value="abc"/>')[0]
+
+    assert type(node) is TextInput
+    assert node.value == 'abc'
+
+    node = HTML('<input type="text" value="xyz" disabled/>')[0]
+
+    assert type(node) is TextInput
+    assert node.value == 'xyz'
+    assert node.disabled == True  # NOQA: E712
```
fscherf, line 351: I assume you want to check if the type of ``node.disable`` is a boolean. If you use ``is`` instead of ``==`` you don't have to add the ``NOQA``

tests/test_0001_html.py
```diff
@@ -336,3 +337,46 @@ def test_html_strings():
...
+    node = HTML('<input type="checkbox"/>')[0]
+
+    assert type(node) is CheckBox
+    assert node.value == False  # NOQA: E712
+
+    node = HTML('<input type="checkbox" checked/>')[0]
+
+    assert type(node) is CheckBox
+    assert node.value == ''
+    # assert node.value == True  # TODO: fix in the next PR
```
fscherf, line 366: Is this broken? This should be an boolean.

fscherf (changes requested):

fscherf:
@maratori Great PR in general! Just some nitpicking

tests/test_0001_html.py
maratori: Fixed

tests/test_0001_html.py
maratori: Good catch, thanks!

tests/test_0001_html.py
maratori: Yes it is broken. I will create PR with fix after this one.

tests/test_0001_html.py
```diff
@@ -336,3 +344,46 @@ def test_html_strings():
...
+    node = HTML('<input value="abc"/>')[0]
+
+    assert type(node) is TextInput
+    assert node.value == 'abc'
+
+    node = HTML('<input type="text" value="xyz" disabled/>')[0]
+
+    assert type(node) is TextInput
+    assert node.value == 'xyz'
+    assert node.disabled is True
```
fscherf, line 358: If you test if inputs can be disabled, you should also test if the previous input ``disabled`` is False

fscherf (changes requested):
Last nitpick. After that i am ok with your changes

tests/test_0001_html.py
maratori: Agree. Fixed.

fscherf (approved):

fscherf:
@maratori do you need a release, or should i wait for your next pr with the ``value`` fixes?

maratori:
@fscherf Let me add few PRs today.

tests/test_0001_html.py
maratori: #84

PR #84, opened 2021-09-12
html: fix CheckBox.value to be always bool

maratori:


lona/html/data_binding/inputs.py
```diff
@@ -127,9 +127,12 @@ class CheckBox(TextInput):
...
-        return self.attributes.get(self.INPUT_ATTRIBUTE_NAME, False)
+    def value(self) -> bool:
+        value = self.attributes.get(self.INPUT_ATTRIBUTE_NAME, False)
+        if value == '':
+            return True
+        return bool(value)
 
     @value.setter
-    def value(self, new_value):
+    def value(self, new_value: bool) -> None:
```
fscherf, line 137: Shouldn't we add an explicit type check like in ``disabled`` here? Or should ``checkbox.value = 'foo'`` be allowed?

maratori: Yeah. Let's do this.
Fixed.

fscherf (approved):

fscherf:
@maratori Great! Thanks! :)

PR #85, opened 2021-09-12
tests: split long tests to class methods

maratori:
Here is why we had a long test functions: https://github.com/lona-web-org/lona/pull/61#discussion_r702608329.

Now we use test class methods in combination with custom `@pytest.mark.incremental()` from [here](https://docs.pytest.org/en/latest/example/simple.html#incremental-testing-test-steps).

The only downside is that we can't mark that test depends on all methods in class, but have to list them all (see RKrahl/pytest-dependency#55).

### DIFF
--- a/tests/test_0001_html.py
+++ b/tests/test_0001_html.py
@@ -2,434 +2,510 @@ import re
 
 import pytest
 
+from lona.html import (
+    NumberInput,
+    TextInput,
+    CheckBox,
+    TextArea,
+    Button,
+    Select,
+    Submit,
+    HTML,
+    Node,
+    Div,
+)
 
-def test_attribute_dict():
-    from lona.html import Div
+
+@pytest.mark.incremental()
+class TestAttributeDict:
+    def test_get_initial_values(self):
+        d = Div(foo='foo', bar='bar')
 
-    # init ####################################################################
-    d = Div(foo='foo', bar='bar')
+        assert d.attributes['foo'] == 'foo'
+        assert d.attributes['bar'] == 'bar'
 
-    assert d.attributes['foo'] == 'foo'
-    assert d.attributes['bar'] == 'bar'
+    def test_set_attributes(self):
+        d = Div()
 
-    # reset ###################################################################
-    d = Div()
+        d.attributes = {
+            'foo': 'foo',
+            'bar': 'bar',
+        }
 
-    d.attributes = {
-        'foo': 'foo',
-        'bar': 'bar',
-    }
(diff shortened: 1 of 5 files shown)


maratori:
@fscherf What do you think about this?

I changed only one test because wasn't sure if you would agree with PR.

fscherf:
@maratori Wow! I like that idea a lot! Can we maybe fully get rid of pytest-dependency with this approach? We could add something like ``FEATURES = ['html-parsing']`` to the test classes and let other tests check if a test with this feature failed before. We would have to name the tests explicitly (``test_feature_0001.py``) but we have to do this with pytest-dependency too.

maratori:
> Can we maybe fully get rid of pytest-dependency with this approach? We could add something like `FEATURES = ['html-parsing']` to the test classes and let other tests check if a test with this feature failed before.

Yes, we can. In a separate PR?

But I'm personally against test dependency at all. I don't think it is a problem if many tests fail at the same time.
As a compromise we can split all tests into suits (unit, integration, e2e, ...). And run next suit only if previous passed.

fscherf:
@maratori That sounds perfekt! A separate pr is a good idea. I will test this pr further and give some feedback. At first glance it looks pretty good

tests/test_0001_html.py
```diff
@@ -2,52 +2,63 @@
...
+    def test_get(self):
+        d = Div(foo='foo', bar='bar')
 
-    assert d.attributes['foo'] == 'foo'
-    assert d.attributes['bar'] == 'bar'
+        assert d.attributes['foo'] == 'foo'
+        assert d.attributes['bar'] == 'bar'
 
-    # reset ###################################################################
-    d = Div()
```
fscherf, line 16: @maratori The test seems to be changed here. I don't think this is critical but, this should be a plain refactoring pr that contains no functional changes

fscherf (changes requested):
@maratori In general i agree with these changes, just some minor nitpicking

tests/test_0001_html.py
maratori: @fscherf Agree. Fixed.

Should I refactor other tests as well in this PR?

fscherf: @maratori Sure, that would be great! :)

maratori: Done

fscherf (approved):

fscherf:
@maratori Great! Thanks for this, looks like a lot of work :)

PR #87, opened 2021-09-12
html: change type of NumberInput.value based on step

maratori:


fscherf:
@maratori wouldn't it be simpler to define a basic type in ``NumberInput``? Something like ``NumberInput(basic_type=int)`` (maybe a better name)

maratori:
@fscherf It could be something like `is_int: bool = False`. But I decided to create a separate class because flag argument is considered as [anti-pattern](https://martinfowler.com/bliki/FlagArgument.html).

fscherf:
@maratori I agree, but passing in a type is no flag. I don't like how much duplication this second class adds, just because one has ``int`` hardcoded and the other ``float``. Essential they do the same.

A compromise could be to set ``BASE_TYPE = float`` in the number class and let it check against this type. The ``IntNumber`` class could inherit from it and override this to int.

Btw: I think ``int`` should be the default type for `NumberInput`` now that i think about it

maratori:
@fscherf Please take a look. Now type of `NumberInput.value` is determined based on `NumberInput.step`.

fscherf:
@maratori Much better! But now ``step`` has two purposes and one is not very obvious. Could we determine the type by the initial value?

``NumberInput(value=1)`` vs ``NumberInput(value=1.03)`` If the initial value is unset, ``int`` is used.

Could you rename ``_T`` to ``Number`` and declare it at the top, before the class definitions?

maratori:
@fscherf I'll answer you later because I realized that is a bit complicated.

Meanwhile I've fixed parsing in #88

fscherf:
@maratori No worries! We are not in a hurry :)

maratori:
@fscherf what do have? There is a number input element in browser with following behaviour:
* User can type any number regardless of `value`/`step`/`min`/`max` attributes. For example: `-123`, `12.3`, or even `-2.3e-5`
* Depending on attributes `value`/`step`/`min`/`max` current value can be valid or not (see css `:invalid` and js `element.validity.valid`)
* Valid value:
  * Can't be less than `min`, if defined
  * Can't be greater than `max`, if defined
  * Is equal to `x + k * step`, where:
    * `k` is any integer number
    * `step` = 1, if not defined
    * `x` = `min` attribute (if defined)
    * or `x` = `value` attribute (if defined)
    * or `x` = 0

For example `<input type="number" value="1.3">` will have valid values `-0.7`, `0.3`, `1.3`, `2.3`, ...

Conclusion 1. My PR is incorrect because it doesn't take into account `min` and `value`.
Conclusion 2. User can type any `float` number into simple `<input type="number">` even if it seems to be `int`.

What behaviour do we expect from `NumberInput` in lona?
1. `NumberInput.value` returns current value from browser as `float` (can be invalid).
1. `NumberInput.value` returns only valid `float`. If current value is not valid, it returns `None`.
1. `NumberInput.value` returns only valid `float`. If current value is not valid, it returns last known valid value.
1. `NumberInput.value` returns only valid `float`. If current value is not valid, it returns the nearest valid value (`round`/`floor`).
1. The same as 2, but it returns `int` if `x` and `step` are integer.
1. The same as 3, but it returns `int` if `x` and `step` are integer.
1. The same as 4, but it returns `int` if `x` and `step` are integer.

fscherf:
@maratori Thanks for elaborating this furter! I will try to merge all current bugfixes, and release 1.7 first. Then i will come back to this.

maratori:
Let's continue in discussion #102

PR #88, opened 2021-09-13
Parse NumberInput

maratori:


lona/html/data_binding/inputs.py
```diff
@@ -173,19 +173,18 @@ def __init__(
         )
         self.min = min
         self.max = max
```
fscherf: Shouldn't we do the same for these to values as well?

fscherf: Oh! That's done in the other commit.

fscherf (changes requested):
Could you add more documentation to the patch descriptions? These patches seem to do multiple things. A sentence on what is broken would help too

maratori:
@fscherf You are right. I messed all in one PR. Will create another.

PR #89, opened 2021-09-13
html: parsing: pass attributes to constructor

maratori:
It will allow to add some validation/conversion inside node constructor.
The only downside - all nodes must accept `attributes` or `kwargs` in constructor.

### DIFF
--- a/lona/html/data_binding/inputs.py
+++ b/lona/html/data_binding/inputs.py
@@ -171,9 +171,15 @@ class NumberInput(TextInput):
             input_delay,
             **kwargs,
         )
-        self.min = min
-        self.max = max
-        self.step = step
+
+        if min is not None:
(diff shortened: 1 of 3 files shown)


maratori:
@fscherf What do you think?

fscherf:
@maratori To be honest: I don't completely understand what You are doing and why this is necessary. Is this infrastructure to fix something for future use?

maratori:
@fscherf I decided that proposed solution is not good enough. Closing PR.
I'll create a discussion about parsing html.

fscherf:
@maratori Ok! I am looking forward towards it. Thanks for Your time :)

maratori:
Just a quick update. I decided that current solution is OK and I will not create a discussion.
Sorry for distraction.

PR #91, opened 2021-09-14
html: fix node list operations

fscherf:
Previously nodes had a special ``__eq__()`` method that made high level comparisons
between nodes possible.

    Div(a=1) == Div(a=1) -> True
    Div(a=1) == Div(b=1) -> False

This feature was intended to be used for testing purposes in the future.

The problem with this implementation is that methods like list.index() or
list.remove() use ``__eq__()`` to find an list item.

    d1 = Div()
    d2 = Div()
    l = [d1, d2]

    l.index(d1) -> 0
    l.index(d2) -> 0

This breaks patch generating and a lot of potential application code.

@maratori Could you review this?

### DIFF
--- a/doc/content/end-user-documentation/views.rst
+++ b/doc/content/end-user-documentation/views.rst
@@ -439,8 +439,13 @@ Input Events
 
 .. note::
 
-    Changed in 1.5: In all versions prior to 1.5, only widgets could handle
+    **Changed in 1.5:** In all versions prior to 1.5, only widgets could handle
     their own events. In versions after 1.5 all node classes can.
+
+    **Changed in 1.7:** In all versions prior to 1.7, ``==`` checked if two
+    nodes have equal attributes, but did not check if node A is the same node
+    as node B.  To check if node A is node B is ``is`` instead of ``==`` was
+    required.
 
 Input events get handled in a chain of hooks. Every hook is required to return
 the given input event, to pass it down the chain, or return ``None`` to mark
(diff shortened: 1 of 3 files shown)


maratori:
@fscherf Don't we need to touch `NodeList.__eq__` as well?

fscherf:
@maratori No i don't think so. To me it seems to be only an issue if a list tries to compare two nodes, not if you compare a list with a NodeList

maratori (approved):

maratori:
@fscherf So from now on can we use `if input_event.node == my_node` instead of `if input_event.node is my_node`?

fscherf:
@maratori oh! Good point! Yes that's true, and we have to change the docs about that

doc/content/end-user-documentation/views.rst
```diff
@@ -439,9 +439,14 @@ Input Events
 
 .. note::
 
-    Changed in 1.5: In all versions prior to 1.5, only widgets could handle
+    **Changed in 1.5:** In all versions prior to 1.5, only widgets could handle
     their own events. In versions after 1.5 all node classes can.
 
+    **Changed in 1.7:** In all versions prior to 1.7, ``==`` checked if two
+    nodes have equal attributes, bot did not check if node A is the same node
```
maratori, line 446:
```suggestion
    nodes have equal attributes, but did not check if node A is the same node
```

fscherf: Oh thanks! Fixed

maratori (approved):

fscherf:
@maratori I decided to release 1.7 because the change behavior in comparisons is a breaking change

PR #94, opened 2021-09-14
html: refactoring: get rid of None in default arguments

maratori:


fscherf:
@maratori Thanks! I have nothing to complain about :)

PR #101, opened 2021-09-15
html: fix parsing of Textarea.value

fscherf:

### DIFF
--- a/lona/html/parsing.py
+++ b/lona/html/parsing.py
@@ -83,6 +83,8 @@ def _setup_node_classes_cache():
 
 
 class NodeHTMLParser(HTMLParser):
+    CDATA_CONTENT_ELEMENTS = HTMLParser.CDATA_CONTENT_ELEMENTS + ('textarea',)
+
     def __init__(self, *args, use_high_level_nodes=True, node_classes=None,
                  **kwargs):
 
@@ -149,9 +151,6 @@ class NodeHTMLParser(HTMLParser):
         self._node.append(node)
         self.set_current_node(node)
 
-    def handle_endtag(self, tag):
-        self.set_current_node(self._node.parent)
-
     def handle_data(self, data):
         text = data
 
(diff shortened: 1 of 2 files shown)


lona/html/parsing.py
```diff
@@ -161,7 +161,11 @@ def handle_data(self, data):
             if not text:
                 return
 
-        self._node.append(TextNode(text))
+        if self._node.tag_name == 'textarea':
+            self._node.value = text
```
maratori, line 165: I'm afraid it is complicated than that. For example following should pass.
```py
assert HTML('<textarea>abc<br>xyz</textarea>')[0].value == 'abc<br>xyz'
assert HTML('<textarea>aaa <b>bbb</b> ccc</textarea>')[0].value == 'aaa <b>bbb</b> ccc'
```

maratori (changes requested):

lona/html/parsing.py
fscherf: @maratori You are right. That's what the javascript does too

fscherf: @maratori Fixed!

lona/html/parsing.py
```diff
@@ -128,6 +153,15 @@ def handle_starttag(self, tag, attrs):
 
             node_attributes[name] = value
 
+        # textareas
+        if self._node.tag_name == 'textarea':
+            self._node.value = self._node.value + self.render_start_tag(
```
maratori, line 158: It seems there is an easier solution. Just add
```py
class NodeHTMLParser(HTMLParser):
    CDATA_CONTENT_ELEMENTS = HTMLParser.CDATA_CONTENT_ELEMENTS + ("textarea",)
```

fscherf: @maratori Oh! Thats much simpler indeed :D Thanks for researching this, this class attribute seems not to be documented

Fixed!

maratori (approved):

fscherf:
@maratori Thanks for reviewing :)

PR #103, opened 2021-09-16
ci: rename github action test to ci

maratori:


fscherf:
Makes sense! Should we rename the file from ``test.yaml`` to ``ci.yaml`` as well?

maratori:
> Makes sense! Should we rename the file from `test.yaml` to `ci.yaml` as well?

Already done.

fscherf:
Oh! I didn't see it in the github diff view. I agree! Thanks :)

PR #104, opened 2021-09-16
doc: readme: add ci badge

maratori:
Works in conjunction with #103


fscherf:
I agree with the changes, but this should be split in two separate commits. One for the new batch and one for the changed alt strings.

maratori:
Fixed

fscherf:
@maratori Much better! Thanks! :)

PR #105, opened 2021-09-17
Use isort to keep imports consistent

maratori:
* add `make isort`
* validate imports in linter
* can't use `flake8-isort` because of gforcada/flake8-isort#101


fscherf:
@maratori Great change! Thanks for that :)

PR #106, opened 2021-09-17
Ignore case while sorting imports

maratori:


fscherf:
@maratori why do we want this exact configuration and not the default one?

maratori:
@fscherf I've discovered that current config is not always use length to sort. Example: https://github.com/lona-web-org/lona/pull/106/files#diff-220e4fd05c64a365ba3a559a3d4113b7382152a54f5902fed428c0ce1603adfaR7

fscherf:
@maratori and why? Is there a good reason why isort is doing it this way by default? Why do we want it differently?

maratori:
@fscherf I've rebased branch and removed configs that don't change defaults.

By default it sorts first by type and then alphabetically (`order_by_type = True` and `length_sort = False`).

When we sort by length, but keep sorting by type, we get following.
```py
from lona.html import (
    TextInput,
    Button,
    Node,
    Div,
    Br,
    A,
    CHANGE,
    CLICK,
    HTML,
)
```

I think it's better to not sort by type.
```py
from lona.html import (
    TextInput,
    CHANGE,
    Button,
    CLICK,
    Node,
    HTML,
    Div,
    Br,
    A,
)
```

fscherf:
@maratori I totally agree! Thanks! :)

PR #107, opened 2021-09-18
dev: add type annotations to view

maratori:


lona/server.py
```diff
@@ -32,6 +40,8 @@
 http_logger = logging.getLogger('lona.server.http')
 websockets_logger = logging.getLogger('lona.server.websockets')
 
+_T = TypeVar('_T')
```
fscherf, line 43: Can we give this a better name?

lona/view.py
```diff
@@ -1,29 +1,90 @@
...
+from lona.request import Request
+
+# avoid import cycles
+if TYPE_CHECKING:
+    from lona.server import LonaServer
+
+
+_T = TypeVar('_T')
+_V = TypeVar('_V', bound='LonaView')
+_HTML = Union[None, AbstractNode, str]
```
fscherf, line 44: Also here. Does it make sense to collect such types in a dedicated module to make the reusable?

lona/view.py
```diff
@@ -1,29 +1,90 @@
-from typing import List, Dict
+from typing import (
```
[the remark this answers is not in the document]
maratori: Yep

lona/server.py
```diff
@@ -1,10 +1,18 @@
...
+from concurrent.futures import Future
 from functools import reduce
 import operator
 import logging
 import inspect
 import asyncio
+import sys
 import os
 
+if sys.version_info >= (3, 8):
```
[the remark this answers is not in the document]
maratori: Let's drop `if` and just import unconditionally
```py
from typing_extensions import Literal
```

lona/server.py
maratori: What about removing underscore?

fscherf: Aren't one character definitions an anti-pattern?

maratori: In this case no. That's how mypy recommends define generics.
https://mypy.readthedocs.io/en/stable/generics.html

maratori: And python docs as well https://docs.python.org/3/library/typing.html#generics

fscherf: @maratori Ah ok. I am ok with that then

lona/server.py
```diff
@@ -1,10 +1,18 @@
...
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
fscherf, line 14: @maratori in the last pr we agreed on just using this import in all python versions. I would do the same here

maratori: It was current pr :)
I pushed the changes just now.

fscherf: @maratori: Whoops :D thanks!

lona/view.py
maratori: I don't think it's a good idea. None of them truly reusable.
`_T` is just any type.
`_V` is a user view and seems is used only in this file.
`_HTML` is also used only here. Different type `Union[AbstractNode, str]` will be used in other files.

lona/server.py
maratori: np :)

maratori: btw: is it necessary to mention you in each comment with @fscherf ?

fscherf:
@maratori the commit message could use some more documentation but in general i am ok with the changes. I have to research some typing things more closely tomorrow (for example i diddn't know ``typing.cast`` before) but everything seems reasonable and clean. Good work! :)

maratori:
> the commit message could use some more documentation

@fscherf What documentation do you mean?

>  I have to research some typing things more closely tomorrow

Sure, take your time.

fscherf:
@maratori I thought it would make sense to document the changed import for ``typing_extentsions`` but i think it's clear enough. I'll take it like this :) Thanks!

PR #109, opened 2021-09-18
lint: use flake8-commas plugin

maratori:
@fscherf What do you think about this [linter](https://github.com/PyCQA/flake8-commas)?


fscherf:
@maratori I love it! I hate that ``a = 2,`` is valid python that creates a tuple. Thanks for that!

PR #110, opened 2021-09-18
tests: add tests for NumberInput

maratori:


tests/test_0001_html.py
```diff
@@ -461,10 +461,10 @@ def test_not_implemented_input_types(self, tp):
         assert type(node) is Node
 
     def test_input_type_number(self):
-        node = HTML('<input type="number" value="123.5"/>')[0]
```
fscherf, line 464: @maratori why is this changed?

maratori: W/o any specific purpose. Reverted.

fscherf:
@maratori Great test! Thanks :)

PR #114, opened 2021-09-19
html: parsing: raise error on unexpected end tag

maratori:


fscherf:
@maratori Great PR! I have nothing to complain about :)

PR #116, opened 2021-09-20
html: fix handling of boolean attributes

fscherf:


fscherf:
@maratori Could you review this? It's the same code we patched before to fix boolean attributes

lona/client/dom-renderer.js
```diff
@@ -49,7 +49,7 @@ Lona.LonaDomRenderer = function(lona_context, lona_window) {
 
                     // properties
                     if(property_names.indexOf(key) > -1) {
-                        if(key != 'value') {
+                        if(key != 'value' && node_attributes[key] === '') {
```
maratori, line 52: I think it is more safe
```suggestion
                        if(key != 'value' && node_attributes[key] !== false) {
```

lona/html/attribute_dict.py
```diff
@@ -1,5 +1,7 @@
 from lona.protocol import PATCH_TYPE, OPERATION
 
+BOOLEAN_ATTRIBUTES = ['checked', 'selected', 'disabled']
```
maratori, line 3: I'm not sure list is complete.

lona/html/attribute_dict.py
```diff
@@ -166,8 +168,10 @@ def to_attribute_string(self, skip_value=False):
                 if skip_value and key == 'value':
                     continue
 
-                if value == '':  # boolean properties
-                    string.append(key)
+                # boolean properties
+                if key in BOOLEAN_ATTRIBUTES:
```
maratori, line 172: I think it's better to fix `CheckBox` instead of patching `AttributeDict`.

lona/html/attribute_dict.py
fscherf: Good point. I didn't found a official list of all boolean attributes, but this list seems to be accurate
https://meiert.com/en/blog/boolean-attributes-of-html/

lona/html/attribute_dict.py
fscherf: My understanding is that these attribute names are boolean on every element type

lona/client/dom-renderer.js
fscherf: @maratori fixed

lona/html/attribute_dict.py
maratori: Seems it almost matches with this list https://html.spec.whatwg.org/#attributes-3

lona/html/attribute_dict.py
maratori: You a right

lona/html/attribute_dict.py
fscherf: @maratori I will use the table from your link

fscherf: @maratori done

maratori (approved):
👍

fscherf:

PR #117, opened 2021-09-20
html: parsing: optional slash in self-closing tags

maratori:
Before this change slash was not optional in self-closing tags.

### DIFF
--- a/tests/test_0001_html.py
+++ b/tests/test_0001_html.py
@@ -441,6 +441,30 @@ class TestHTMLFromStr:
         ):
             HTML('</span>')
 
+    def test_self_closing_tag_with_slash(self):
+        img = HTML('<div><img src="abc"/></div>')[0].nodes[0]
+
+        assert img.tag_name == 'img'
+        assert img.self_closing_tag is True
+
+    def test_self_closing_tag_without_slash(self):
+        img = HTML('<div><img src="abc"></div>')[0].nodes[0]
+
+        assert img.tag_name == 'img'
+        assert img.self_closing_tag is True
+
+    def test_non_self_closing_tag(self):
+        div = HTML('<div></div>')[0]
+
+        assert div.tag_name == 'div'
+        assert div.self_closing_tag is False
+
+    def test_non_self_closing_tag_with_slash(self):
+        span = HTML('<div><span/></div>')[0].nodes[0]
+
+        assert span.tag_name == 'span'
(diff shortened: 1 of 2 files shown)


tests/test_0001_html.py
```diff
@@ -441,6 +441,21 @@ def test_missing_start_tag(self):
...
+        assert node.nodes[0].tag_name == 'img'
+        assert node.nodes[0].self_closing_tag is True
+
+    def test_self_closing_tag_without_slash(self):
+        node = HTML('<div><img src="abc"></div>')[0]
+        assert node.nodes[0].tag_name == 'img'
+        assert node.nodes[0].self_closing_tag is True
+
+    def test_non_self_closing_tag_with_slash(self):
+        node = HTML('<div><span/></div>')[0]
```
fscherf, line 455: Should this work? That's no valid html.

maratori: I thought about that. But decided not to change current behavior.

`<div><span/></div>` indeed invalid html5.
But this `<svg xmlns="http://www.w3.org/2000/svg"><path/></svg>` is valid html.
There are [foreign elements](https://html.spec.whatwg.org/#foreign-elements) that can be self closing in html5.

So I decided not to add rules to determine if tag can be self closing.
We can do it in a separate PR, if necessary.

maratori: What means "check this line"?

fscherf: @maratori I am sorry. ``handle_starttag`` now has a keyword argument ``self_closing_tag``. Don't we have to check the value of ``self_closing_tag`` in https://github.com/lona-web-org/lona/blob/master/lona/html/parsing.py#L137 ?

maratori: It depends on what behavior we expect here
```py
        node = HTML('<div><span/></div>')[0]
        assert node.nodes[0].tag_name == 'span'
        assert node.nodes[0].self_closing_tag is False  # or True
```
I'd say that we should not.

fscherf: I think an argument to let ``self_closing_tag`` be true in this case would be that ``str(node)`` would generate the same (broken) HTML that was parsed.

maratori: Fixed

fscherf:
@maratori Great changes! Thanks for that. I will release 1.7.1 this day.

PR #118, opened 2021-09-20
html: fix error message in AttributeList

maratori:


fscherf:
@maratori Great! Thanks :)

PR #121, opened 2021-09-21
lint: use flake8-length plugin

maratori:
[flake8-length](https://github.com/orsinium-labs/flake8-length) is less strict than default flake8 rules E501 and W505 about line length. Default for `max-line-length` option is still 79.


fscherf:
@maratori Ouh, thats great! Thanks for that!

PR #122, opened 2021-09-21
lint: use flake8-mutable plugin

maratori:
Let's use [flake8-mutable](https://github.com/ebeweber/flake8-mutable).


fscherf:
@maratori Great idea! Mutable default arguments are a stable source of bugs :D

PR #123, opened 2021-09-21
lint: use flake8-use-fstring plugin

maratori:
Let's use [flake8-use-fstring](https://github.com/MichaelKim0407/flake8-use-fstring)

### DIFF
--- a/flake8.ini
+++ b/flake8.ini
@@ -17,5 +17,8 @@ per-file-ignores =
 # require code(s) to be included in # noqa comments
 noqa-require-code = True
 
+# flake8-use-fstring: complain about any usage of %-style string formatting
+percent-greedy = 2
+
 # enable flake8-logging-format
 enable-extensions=G
--- a/flake8.ini
+++ b/flake8.ini
@@ -14,11 +14,12 @@ per-file-ignores =
     # imported but unused
     __init__.py: F401
 
-# require code(s) to be included in # noqa comments
+# flake8-noqa: require code(s) to be included in # noqa comments
 noqa-require-code = True
 
 # flake8-use-fstring: complain about any usage of %-style string formatting
 percent-greedy = 2
 
-# enable flake8-logging-format
-enable-extensions=G
+enable-extensions =
+    # enable flake8-logging-format
+    G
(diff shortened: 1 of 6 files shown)


lona/response_parser.py
```diff
@@ -50,7 +50,7 @@ def render_response_dict(self, raw_response_dict, view_name):
             # check keys
             if len(set(raw_response_dict) & key_words) != 1:
                 raise RuntimeError(
-                    'response dicts have to contain exactly one keyword of: {}'.format(  # NOQA: LN001
+                    'response dicts have to contain exactly one keyword of: {}'.format(  # NOQA: LN001,FS002
```
fscherf, line 53: Why do we leave this line like this?

lona/static_files.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, name, path, url=None, sort_order=None, link=True,
         self.static_url_prefix = ''
 
     def __repr__(self):
-        return '<{}({}, sort_order={}, enabled={}, {}{} -> {})>'.format(
+        return '<{}({}, sort_order={}, enabled={}, {}{} -> {})>'.format(  # NOQA: FS002
```
fscherf, line 27: Why do we leave this line like this?

lona/response_parser.py
maratori: While migrating to f-strings in #58 I decided that in some cases `.format` is more readable.
All such cases now have `# NOQA`.
If you think it's better to rewrite them using f-strings, I can update PR.

flake8.ini
```diff
@@ -14,8 +14,12 @@ per-file-ignores =
     # imported but unused
     __init__.py: F401
 
-# require code(s) to be included in # noqa comments
+# flake8-noqa: require code(s) to be included in # noqa comments
```
fscherf, line 17: This should be it's own commit, or the change should be at least mentioned in the commit message

fscherf (changes requested):

flake8.ini
maratori: Fixed

fscherf (approved):
Great! Thanks! :)

PR #124, opened 2021-09-21
lint: use flake8-2020 plugin

maratori:
Let's use [flake8-2020](https://github.com/asottile/flake8-2020) to check for misuse of `sys.version` or `sys.version_info`. May be important after Python 3.10.


fscherf:
@maratori Great idea! Thanks for adding this!

PR #126, opened 2021-09-21
dev: close file after read

maratori:


fscherf:
@maratori Great! Nothing to complain, just one question: Why have some of your commits the prefix "dev: "?

fscherf:
@maratori I knew "dev: " as a prefix for commits that are not supposed to be upstreamed but for changes that make local development work (but i can't remember where i got this from)

I think it's a good idea to document a list of prefixes! For now we can agree on your definition.

maratori:
@fscherf I was 100% sure that I saw this prefix in your commits. But now I can't find it :)

fscherf:
@maratori thanks! :)

PR #127, opened 2021-09-22
scripts: add command line args

fscherf:
With these patches command line arguments can be used in Lona scripts.

``python lona_script.py --port=8081 --shell -O Foo=True``

### DIFF
--- a/lona/app.py
+++ b/lona/app.py
@@ -82,6 +82,7 @@ class LonaApp:
             http_pass_through: bool = False,
             frontend_view: Union[None, str, LonaView] = None,
     ) -> Callable[[Type[LonaView]], None]:
+
         def decorator(view_class: Type[LonaView]) -> None:
             self.routes.append(
                 Route(
@@ -96,6 +97,7 @@ class LonaApp:
 
         return decorator
 
+    # middleware
     @overload
     def middleware(self) -> Callable[[Type], None]:
         ...
@@ -110,11 +112,13 @@ class LonaApp:
 
         if callable(middleware_class):
             decorator(middleware_class)
+
             return None
 
         else:
             return decorator
 
+    # frontend
     @overload
     def frontend_view(self) -> Callable[[Type[LonaView]], None]:
         ...
@@ -129,6 +133,7 @@ class LonaApp:
 
         if callable(view_class):
             decorator(view_class)
+
             return None
 
         else:
@@ -142,6 +147,7 @@ class LonaApp:
             string: str = '',
             path: str = '',
     ) -> None:
+
         if name.startswith('/'):
             name = name[1:]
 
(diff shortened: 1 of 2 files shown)


lona/app.py
```diff
@@ -240,13 +333,12 @@ def setup_server(
     def run(
             self,
             loop: Union[AbstractEventLoop, None] = None,
+            parse_command_line: int = True,
```
maratori, line 336:
```suggestion
            parse_command_line: bool = True,
```

lona/app.py
```diff
@@ -202,17 +201,105 @@ def add_static_file(
...
+                args['settings_post_overrides'],
+            )
+
+        return args
+
     # server ##################################################################
     def setup_server(
             self,
             loop: Union[AbstractEventLoop, None] = None,
+            settings_pre_overrides: Union[Dict, None] = None,
```
maratori, line 301: nit:
```suggestion
            settings_pre_overrides: Optional[Dict[str, Any]] = None,
```

maratori (changes requested):
I believe parsing command line arguments is not a responsibility of `LonaApp`.
It already can receive arbitrary arguments in `app.run()`.

If I want to parse arguments I should do it outside of `LonaApp`.
We can add some helper function. For example
```py
app = LonaApp(__file__)
# setup views
app.run(**parse_command_line(port=8080))
```

lona/app.py
fscherf: Fixed

lona/app.py
fscherf: Fixed

lona/app.py
```diff
@@ -257,10 +349,32 @@ def run(
...
         setup_logging(server_args)
 
+        # setup server
+        self.setup_server(
+            loop=loop,
+            settings_pre_overrides=getattr(
+                server_args,
+                'settings_pre_overrides',
+                None,
+            ),
```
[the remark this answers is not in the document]
fscherf: fixed

fscherf:
@maratori I think it's a very useful feature that you just call ``app.run()`` and can override the port for example. If we get a bug report with a reproducing script, the first thing i do is patching ``app.run()`` to ``app.run(log_level='debug', shell=True)``. It would be nice to just run it with ``-l debug --shell`` without changing something.

If we put this functionality in it's own place, i would recommend everyone to use it, because it adds the same interface to scripts like in projects, which is especially useful if you read https://lona-web.org/end-user-documentation/debugging.html while developing. I think it makes it consistent.

Also: I like debugging infrastructure that's always in place. Not only if you set it up. I can tell every user to add ``-l debug``, but not every user to patch a line of code.

Is there a reason to split that, that i miss?

maratori (approved):
@fscherf Ok

fscherf:

themutt (approved):

PR #128, opened 2021-09-23
html: data binding: CheckBox: remove obsolete argument "input_delay"

fscherf:
input_delay makes only sense with text inputs.


maratori (approved):

fscherf:

PR #133, opened 2021-09-26
lint: change minimal python version in mypy config

maratori:


fscherf:
@maratori Great! Thanks! :)

PR #134, opened 2021-09-26
Define `STATIC_FILES` in `LonaView` like in `Widget`

maratori:
Closes #95

* A lot of refactoring
* Some minor fixes
* Add logging on non-unique names
* Discover static files in views
* Add tests

### DIFF
--- a/lona/html/abstract_node.py
+++ b/lona/html/abstract_node.py
@@ -1,4 +1,7 @@
+from typing import List
+
 from lona.unique_ids import generate_unique_id
+from lona.static_files import StaticFile
 from lona.html.selector import Selector
 
 
@@ -20,6 +23,8 @@ class DummyDocument:
 
 
 class AbstractNode:
+    STATIC_FILES: List[StaticFile] = []
+
     def __copy__(self, *args, **kwargs):
         raise RuntimeError('copy is not supported')
 
--- a/lona/static_file_loader.py
+++ b/lona/static_file_loader.py
@@ -123,11 +123,9 @@ class StaticFileLoader:
                 discovered_names.append(static_file.name)
 
         # sort
-        self.node_stylesheets = sorted(
-            self.node_stylesheets, key=lambda x: x.sort_order.value)
-
-        self.node_scripts = sorted(
-            self.node_scripts, key=lambda x: x.sort_order.value)
+        self.node_stylesheets.sort(key=lambda x: x.sort_order.value)
+        self.node_scripts.sort(key=lambda x: x.sort_order.value)
+        self.node_static_files.sort(key=lambda x: x.sort_order.value)
 
         # enable or disable static files
         self.static_files = [
--- a/lona/static_file_loader.py
+++ b/lona/static_file_loader.py
@@ -44,83 +44,20 @@ class StaticFileLoader:
         visited_declarations: List[StaticFile] = []
 
         for node_class in self.node_classes:
+            node_class_path = get_file(node_class)
+            node_class_dirname = os.path.dirname(node_class_path)
+            context = f'{node_class_path}::{node_class.__name__}'
             for file_declaration in node_class.STATIC_FILES:
-                node_class_path = get_file(node_class)
-                node_class_dirname = os.path.dirname(node_class_path)
+                static_file = self._process_declaration(
+                    file_declaration,
+                    context,
+                    node_class_dirname,
+                    visited_declarations,
+                    discovered_names,
+                )
 
-                # check static file
-                if not isinstance(file_declaration, StaticFile):
-                    logger.error(
-                        '%s::%s: unknown type found: %r',
-                        node_class_path,
-                        node_class.__name__,
-                        file_declaration,
-                    )
-
+                if static_file is None:
                     continue
-
-                if file_declaration in visited_declarations:
-                    continue
-
-                visited_declarations.append(file_declaration)
-
-                if file_declaration.name in discovered_names:
-                    logger.error(
-                        "%s::%s: static file with name '%s' already used",
-                        node_class_path,
-                        node_class.__name__,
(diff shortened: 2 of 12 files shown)


maratori:
@fscherf I'll rebase this PR after #136 is merged.

fscherf:
@maratori Could you add documentation for the new feature?

maratori:
@fscherf Done.

But documentation is not my best skill. So feel free to correct it.

maratori:
@fscherf CI hangs on python 3.7. I don't know why.

I finished my research on following.
CI passes on master branch now.
I created [branch](https://github.com/maratori/lona/tree/fail) from master with minimal change to [hang CI](https://github.com/maratori/lona/actions/runs/1286943887) like it happened in this PR.

fscherf:
@maratori strange. I noticed that the ci won't finish this morning and restartet the run, but it didn't help. @sobolevn any idea what's causing this?

maratori:
@fscherf I split test, now ci is green

fscherf:
@maratori ok!? Is this a playwright problem? It confuses me that this only happens on 3.7 and is fixed by something that looks unrelated to me

maratori:
@fscherf I don't know what is that. Just found out a workaround.

lona/static_file_loader.py
```diff
@@ -228,3 +214,15 @@ def resolve_path(self, path: str) -> Optional[str]:
         logger.debug("'%s' was not found", path)
 
         return None
+
```
fscherf, line 306: Why are there two functions with new naming outside of the class now? Also why is the first function named class_tree but returns a list of types?

lona/static_file_loader.py
```diff
@@ -1,9 +1,20 @@
-from typing import TYPE_CHECKING, Optional, TypeVar, Type, List, Set
```
fscherf: I don't understand why this is necessary. I understand why to use dataclasses, the rest looks like it makes things nore complicated

tests/test_static_files.py
```diff
@@ -8,16 +8,8 @@
 
 
 def setup_app(app):
-    @app.route('/')
```
fscherf, line 11: So there is no problem here like you suspected in #136 ?

fscherf (changes requested):
@maratori I am ok with these changes in general, but i find some of your changes a little bit complicated

tests/test_static_files.py
maratori: I don't know why I saw failed tests. I can't reproduce it now. So I didn't create an issue and just fixed test. Commit has short description what happened.

lona/html/abstract_node.py
```diff
@@ -1,4 +1,7 @@
+from typing import List
```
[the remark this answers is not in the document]
maratori: Fixed

lona/static_file_loader.py
maratori: Fixed by removing `class_tree` function

tests/test_static_files.py
fscherf: Could you add something like this comment to the commit message? "By mistake" is not much information in case someone is wondering about this commit in a year

lona/static_file_loader.py
maratori: At the first sight it may seem more complicated, but I don't think so.

Actually there are two different objects: declaration of the file and real file (with correct attributes).
The same class was used in both cases. So it was necessary to `copy` and mutate attributes.
That was very unobvious and hidden deep inside the code.

Now they are explicitly split. And it is clear which class has which area of responsibility. And which class has which attributes.

tests/test_static_files.py
maratori: Fixed

lona/static_file_loader.py
fscherf: Ahhhh! Now i get it. Some explanation like this comment should be in the commit message.

lona/static_file_loader.py
fscherf: The naming is still odd. And why has it to be outside the class now?

maratori: > The naming is still odd.

Are you talking about `def all_subclasses`? What's wrong with it?

> And why has it to be outside the class now?

I'd ask why it should be inside class if it isn't related to class at all? It's a separate (and pure) function which can be used in any class/function.

fscherf: All other functions like that in the project start with ``get_``.

Yes its not inherently related to the class, but if we decide to make it an standalone helper it shouldn't be stored in this module. Maybe ``lona.imports``?

lona/static_file_loader.py
maratori: Agree, added explanation to commit message.

lona/static_file_loader.py
maratori: Good point. Fixed.

lona/static_file_loader.py
fscherf: I thought a lot about that and reread the code. I am sorry but i don't like this approach. From a plain modeling standpoint it makes sense, but it is a new abstraction with new types, new helper functions etc, to save one ``copy()``. Also ``StaticFile`` is very basic type. It's basically a dict with a nice name, so user code gets more verbose. I think, it's not worth the complexity.

I like the use of ``dataclass``, so i want to keep this. And i agree, it can be confusing that this copy is necessary, but i would want to fix that with a simple comment

maratori: Ok, I'll merge them back.

Note: user's code will not change at all. There are no breaking changes.

maratori: Done

fscherf:
@maratori Great PR and a great feature! Thanks for that, and thanks for your patience :)

PR #136, opened 2021-09-27
fix(static files): issue when non-linked file was not served

### LINKED ISSUE #135: Static file is not served if `link=False`
After #131 lona don't add static files with `link=False` to HTML. But also it stopped serving them.

maratori:
Fixes #135

### DIFF
--- a/lona/static_file_loader.py
+++ b/lona/static_file_loader.py
@@ -146,10 +146,6 @@ class StaticFileLoader:
             if static_file.name in self.server.settings.STATIC_FILES_DISABLED:
                 static_file.enabled = False
 
-            # disable static files that are flaged as to be not linked
-            if not static_file.link:
-                static_file.enabled = False
-
     def discover(self):
         logger.debug('discover node classes')
 
@@ -170,7 +166,8 @@ class StaticFileLoader:
         self.style_tags_html = self.server.templating_engine.render_template(
             self.server.settings.STATIC_FILES_STYLE_TAGS_TEMPLATE,
             {
-                'stylesheets': [i for i in self.node_stylesheets if i.enabled],
+                'stylesheets': [i for i in self.node_stylesheets
+                                if i.enabled and i.link],
             },
         )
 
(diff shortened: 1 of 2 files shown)


lona/static_file_loader.py
```diff
@@ -170,15 +166,15 @@ def discover(self):
         self.style_tags_html = self.server.templating_engine.render_template(
             self.server.settings.STATIC_FILES_STYLE_TAGS_TEMPLATE,
             {
-                'stylesheets': [i for i in self.node_stylesheets if i.enabled],
+                'stylesheets': [i for i in self.node_stylesheets if i.enabled and i.link],  # NOQA: LN001
```
fscherf, line 169: Could you put this in a separate line or add a newline?

tests/test_static_files.py
```diff
@@ -0,0 +1,53 @@
...
+    @app.route('/')
+    class MyLonaView(LonaView):
+        def handle_request(self, request):
+            return 'SUCCESS'
+
+
+async def test_static_files(lona_app_context):
+    some_existing_file = os.path.basename(__file__)
+
+    # isn't discovered in setup_app()
```
fscherf, line 20: What does this mean?

fscherf (changes requested):
@maratori Great changes! In general i am ok with this, just some nitpicking

Thanks for adding tests! :)

tests/test_static_files.py
maratori: Would it be more clear if I change comment to this?
```py
# if you put widget into setup_app(), lona will not discover static files
```

fscherf: Oh! That sounds like a bug. We should add a "FIXME: " then

maratori: I've fixed comment.

I'm not sure it's a bug. However need to investigate for sure.
I'll create an issue for that.

lona/static_file_loader.py
maratori: Fixed

tests/test_static_files.py
fscherf: Let's add the fixme anyway. The widget should be part of the setup stage, not the test i think

maratori: Added

fscherf:
@maratori Thanks! :)

PR #137, opened 2021-09-30
Add view events for better multi-user views

fscherf:
This pull-request adds a new events that can be send between views for communication. More details: #79

### DIFF
--- a/lona/app.py
+++ b/lona/app.py
@@ -236,7 +236,7 @@ class LonaApp:
 
         parser.add_argument(
             '--debug-mode',
-            choices=['messages', 'views', 'input-events'],
+            choices=['messages', 'views', 'input-events', 'view-events'],
         )
 
         parser.add_argument(
--- a/lona/events/view_event.py
+++ b/lona/events/view_event.py
@@ -1,0 +1,15 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from typing_extensions import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from lona.view import LonaView
+
+
+@dataclass
+class ViewEvent:
+    name: str
+    data: dict
+    view_classes: list[type[LonaView]]
--- a/lona/server.py
+++ b/lona/server.py
@@ -29,6 +29,7 @@ from lona.connection import Connection
 from lona.settings import Settings
 from lona.protocol import METHOD
 from lona.routing import Router
+from lona.view import LonaView
 
 DEFAULT_SETTINGS = os.path.join(
     os.path.dirname(__file__),
@@ -597,3 +598,16 @@ class LonaServer:
 
     def reverse(self, *args, **kwargs):
         return self.router.reverse(*args, **kwargs)
+
+    def fire_view_event(
+            self,
+            name: str,
+            data: dict | None = None,
+            view_classes: type[LonaView] | list[type[LonaView]] | None = None,
+    ) -> None:
+
+        self.view_runtime_controller.fire_view_event(
+            name=name,
+            data=data,
+            view_classes=view_classes,
+        )
--- a/lona/view.py
+++ b/lona/view.py
@@ -21,8 +21,8 @@ from lona.request import Request
 
 # avoid import cycles
 if TYPE_CHECKING:
+    from lona.events.view_event import ViewEvent
     from lona.server import LonaServer
-
 
 T = TypeVar('T')
 V = TypeVar('V', bound='LonaView')
@@ -309,6 +309,14 @@ class LonaView:
             html=html,
         )
 
+    # view events #############################################################
+    def fire_view_event(self, name: str, data: dict | None = None) -> None:
+        self.server.view_runtime_controller.fire_view_event(
+            name=name,
+            data=data,
+            view_classes=[self.__class__],
+        )
+
     # runtime #################################################################
     @overload
     def sleep(self, delay: float) -> None:
@@ -355,6 +363,9 @@ class LonaView:
 
     def handle_input_event(self, input_event: InputEvent) -> None | InputEvent:
         return input_event
+
+    def on_view_event(self, view_event: 'ViewEvent') -> None:
+        pass
 
     def on_shutdown(
             self,
--- a/lona/view_runtime_controller.py
+++ b/lona/view_runtime_controller.py
@@ -1,13 +1,18 @@
+from __future__ import annotations
+
 import contextlib
 import logging
 
 from yarl import URL
 
 from lona.protocol import encode_http_redirect, METHOD
+from lona.events.view_event import ViewEvent
 from lona.view_runtime import ViewRuntime
 from lona.exceptions import ServerStop
+from lona.view import LonaView
 
 input_events_logger = logging.getLogger('lona.input_events')
+view_events_logger = logging.getLogger('lona.view_events')
 views_logger = logging.getLogger('lona.views')
 
 
@@ -314,4 +319,78 @@ class ViewRuntimeController:
                 view_runtime_id=view_runtime_id,
                 method=method,
                 payload=payload,
+            )
+
+    # events ##################################################################
+    def run_view_event_hook(
+            self,
+            view_runtime: ViewRuntime,
+            view_event: ViewEvent,
+    ) -> None:
+
+        try:
+            view_events_logger.debug(
+                'running %r with %r',
+                view_runtime.view.on_view_event,
+                view_event,
+            )
+
+            view_runtime.view.on_view_event(view_event)
+
+        except Exception:
+            view_events_logger.exception(
+                'Exception raised while running %r',
+                view_runtime.view.on_view_event,
+            )
+
+            return
+
+        # send html updates if necessary
+        if view_runtime.document.is_dirty:
+            view_runtime.view.show()
+
+    def fire_view_event(
+            self,
+            name: str,
+            data: dict | None = None,
+            view_classes: type[LonaView] | list[type[LonaView]] | None = None,
+    ) -> None:
+
+        data = data or {}
+        view_classes = view_classes or []
+
+        if not isinstance(view_classes, list):
+            view_classes = [view_classes]
+
+        view_events_logger.debug(
+            'sending ViewEvent(name=%r, data=%r, view_classes=%r)',
+            name,
+            data,
+            view_classes,
+        )
+
+        for view_runtime in self.iter_view_runtimes():
+            if view_classes and view_runtime.view_class not in view_classes:
+                continue
+
+            # if on_view_event is not defined by the view class
+            # it is unnecessary to start a thread
+            if view_runtime.view_class.on_view_event == LonaView.on_view_event:
+                view_events_logger.debug(
+                    '%r skipped (view class does not implement on_view_event())',
+                    view_runtime.view,
+                )
+
+                continue
+
+            view_event = ViewEvent(
+                name=name,
+                data=data.copy(),
+                view_classes=view_classes.copy(),
+            )
+
+            self.server.run_function_async(
+                self.run_view_event_hook,
+                view_runtime,
+                view_event,
             )
--- a/tests/test_view_events.py
+++ b/tests/test_view_events.py
@@ -1,0 +1,98 @@
+from lona.pytest import eventually
+from lona import LonaView
+
+
+def setup_app(app):
+
+    @app.route('/view-1/')
+    class View1(LonaView):
+        def handle_request(self, request):
+            if 'fire' in request.GET:
+                self.fire_view_event('view1', {'view1': True})
+
+        def on_view_event(self, view_event):
+            self.server.state['view1'].append(
+                (view_event.name, view_event.data),
+            )
+
+    @app.route('/view-2/')
+    class View2(LonaView):
+        def handle_request(self, request):
+            if 'fire' in request.GET:
+                self.fire_view_event('view2', {'view2': True})
+
+        def on_view_event(self, view_event):
+            self.server.state['view2'].append(
+                (view_event.name, view_event.data),
+            )
+
+    @app.route('/broadcast/')
+    class FireBroadcastEvent(LonaView):
+        def handle_request(self, request):
+            self.server.fire_view_event('broadcast', {'broadcast': True})
+
+
+async def test_view_events(lona_app_context):
+    """
+    This test tests view events by creating two views that send events to all
+    of their instances, and one view that sends events to all views in the
+    system.
+    The test is successful if both of the first two views only received their
+    own events, and the broadcast events, and no event from view 1 leaked to
+    view 2 or vice versa.
+    """
+
+    from playwright.async_api import async_playwright
+
+    context = await lona_app_context(setup_app)
+
+    async with async_playwright() as playwright:
+        context.server.state['view1'] = []
+        context.server.state['view2'] = []
+
+        browser = await playwright.chromium.launch()
+        browser_context = await browser.new_context()
+
+        # view 1
+        view1_page1 = await browser_context.new_page()
+        view1_page2 = await browser_context.new_page()
+        view1_page3 = await browser_context.new_page()
+
+        await view1_page1.goto(context.make_url('/view-1/'))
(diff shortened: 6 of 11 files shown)


fscherf:
@maratori hm, i seem to have i similar ci problem like you in #134, but i can't split my test into multiple files

sobolevn:
@fscherf do you mean the running duration problem?

fscherf:
@sobolevn Yes. On 3.7 it won't finish or crash. but only on 3.7

sobolevn:
Does the same happen on your local machine?

fscherf:
@sobolevn No. I was not able to reproduce this on any other system but the ci

sobolevn:
You can try to add https://pypi.org/project/pytest-timeout/
This will help to find which tests are causing this.

fscherf:
@sobolevn Good idea. But is there no way to get some output before ``tox `` exits?

sobolevn:
Output itself might be a problem.

fscherf:
@sobolevn I've setup pytest-timeout and now everything is green. Great! -.-

maratori:
@fscherf it's very weird. I did research.

1. Without `pytest-timeout` ci is either green ([branch](https://github.com/maratori/lona/tree/ci-pass), [logs](https://github.com/maratori/lona/actions/runs/1292263039)), or **hangs** ([branch](https://github.com/maratori/lona/tree/ci-hang), [logs](https://github.com/maratori/lona/actions/runs/1292275374))
2. With `pytest-timeout` ci is either green, or **fails** ([branch](https://github.com/maratori/lona/tree/ci-fail), [logs](https://github.com/maratori/lona/actions/runs/1292294950))
3. If ci fails (with `pytest-timeout`), it fails due to static files test failure. Lona doesn't find all `Node` subclasses.
4. Don't know why it hans (without `pytest-timeout`), but most probably the issue is the same - static files discovery.
5. I can't find any correlation when it is green and when hangs/fails.

At least now with `pytest-timeout` we can see what's happening. And another good point is that it is stable (we can reproduce this).

maratori:
Wow, even with `@pytest.mak.skip` ci may hang ([branch](https://github.com/maratori/lona/tree/ci-hang-with-skip), [logs](https://github.com/maratori/lona/actions/runs/1292738706)).

tests/test_view_events.py
```diff
@@ -0,0 +1,100 @@
...
+        await view2_page2.goto(context.make_url('/view-2/'))
+        await view2_page3.goto(context.make_url('/view-2/?fire'))
+
+        # broadcast
+        broadcast_page = await browser_context.new_page()
+
+        await broadcast_page.goto(context.make_url('/broadcast/'))
+
+        # wait for all threads to complete
+        await asyncio.sleep(1)
```
maratori, line 81: I think it's better to avoid `sleep`s in tests.
For example I used in my projects following helper https://gist.github.com/maratori/118d1018893b367be8d35ccda996d3af.

maratori (commented):
@fscherf I like this PR a lot. It's very useful in my project!

Some non-critical comments from my side.

lona/events/view_event.py
```diff
@@ -0,0 +1,24 @@
+from typing import TYPE_CHECKING, Optional, TypeVar, Union, Type, List, Dict
+
+if TYPE_CHECKING:
+    from lona.view import LonaView
+
+T = TypeVar('T', bound='LonaView')
+
+
+class ViewEvent:
```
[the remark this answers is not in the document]
fscherf: Good idea!

lona/events/view_event.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(
+            self,
+            name: str,
+            data: Optional[Dict] = None,
+            view_classes: Optional[Union[Type[T], List[Type[T]]]] = None,
+            issuer: Optional[Union[T, str]] = None,
```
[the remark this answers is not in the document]
fscherf: Hm. The word "Optional" makes much sense to me in this context. Is there a conventional for this? I am ok with both

lona/events/view_event.py
```diff
@@ -0,0 +1,24 @@
+from typing import TYPE_CHECKING, Optional, TypeVar, Union, Type, List, Dict
+
+if TYPE_CHECKING:
+    from lona.view import LonaView
+
+T = TypeVar('T', bound='LonaView')
```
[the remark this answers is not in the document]
fscherf: Ah you are right! I was worried about circular imports, but that seems to be no problem here.  I will fix that

lona/events/view_event.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class ViewEvent:
+    def __init__(
+            self,
+            name: str,
+            data: Optional[Dict] = None,
+            view_classes: Optional[Union[Type[T], List[Type[T]]]] = None,
```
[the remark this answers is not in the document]
fscherf: Ah! Because ``fire_event`` always makes a list of it before. I will fix that

lona/view.py
```diff
@@ -317,6 +319,15 @@ def await_change(self, *nodes, html=None):
             html=html,
         )
 
+    # view events #############################################################
+    def fire_view_event(self, name: str, data: Optional[Dict]) -> None:
+        self.server.view_runtime_controller.fire_event(
```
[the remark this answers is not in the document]
fscherf: The idea was that the call is part of the server, and that it really is part of the view runtime controller is just an implementation detail, but maybe thats not really necessary.

lona/view.py
```diff
@@ -317,6 +319,15 @@ def await_change(self, *nodes, html=None):
             html=html,
         )
 
+    # view events #############################################################
+    def fire_view_event(self, name: str, data: Optional[Dict]) -> None:
```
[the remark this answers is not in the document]
fscherf: I thought of this too, as soon as is pull-requested. Yes, lets make that optional!

lona/server.py
```diff
@@ -594,3 +594,6 @@ def get_view_class(self, route=None, import_string=None, url=None):
 
     def reverse(self, *args, **kwargs):
         return self.router.reverse(*args, **kwargs)
+
+    def fire_view_event(self, *args, **kwargs):
```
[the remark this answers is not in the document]
fscherf: Good point! I will fix that

tests/test_view_events.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # check captured events
+        assert list(context.server.state['view1']) == [
+            ('view1', {'view1': True}),
+            ('view1', {'view1': True}),
+            ('view1', {'view1': True}),
+            ('broadcast', {'broadcast': True}),
```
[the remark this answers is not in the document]
fscherf: Good point! So we should sort the list before checking it.

lona/view_runtime_controller.py
```diff
@@ -315,3 +319,61 @@ def handle_lona_message(self, connection, window_id, view_runtime_id,
...
+        # send html updates if necessary
+        if view._view_runtime.document.is_dirty:
+            view.show()
+
+    def fire_event(
+            self,
+            name: str,
+            data: Optional[Dict] = None,
+            view_classes: Optional[Union[Type[LonaView], List[Type[LonaView]]]] = None,  # NOQA: LN001
+            issuer: Optional[Union[LonaView, str]] = None,
```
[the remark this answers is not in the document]
fscherf: The idea was that if you have views that have special roles, like one is the moderater of a chat, and all others are participants, you could set the issuer to 'moderator'. But when i think about it now, i think the whole field is unnecessary. If you need that, you can add these kind of information to the event data instead. For debugging, the issuer should just be logged

lona/view_runtime_controller.py
```diff
@@ -1,13 +1,17 @@
...
 from yarl import URL
 
 from lona.protocol import encode_http_redirect, METHOD
+from lona.events.view_event import ViewEvent
 from lona.view_runtime import ViewRuntime
 from lona.exceptions import ServerStop
+from lona.view import LonaView
 
 input_events_logger = logging.getLogger('lona.input_events')
+view_events_logger = logging.getLogger('lona.view_events')
```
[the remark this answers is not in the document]
fscherf: Yes! I will fix that

tests/test_view_events.py
fscherf: I am not really happy with this either, but we cant just wait for the call to return, because ``fire_event`` schedules the callbacks to the worker threads anf returnes immediately

fscherf:
@maratori I get the suspicious that this is a asyncio+python3.7 problem is. That might explain why even skips don't work. I tried to reproduce this on multiple systems, but I had no luck.

lona/events/view_event.py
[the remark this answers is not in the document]
maratori: 1. I couldn't find any recommendations what is preferred.
2. There is a new syntax for type annotations. It allows to migrate from `t.Union[int, t.List[str]]` to `int | list[str]`. See [PEP 604](https://www.python.org/dev/peps/pep-0604/) and [PEP 585](https://www.python.org/dev/peps/pep-0585/).
3. I've just found out that it can be used in older python versions with `import __future__` (see [example](https://github.com/maratori/lona/commit/b5a8968f0c25d5d955e2bea9b46fd153547814f9)).
4. If you don't mind, I'd like to use new syntax in the project (I'll prepare PR). Any objections?
5. Going back to original question :) We have two options `None | T | str` vs `Optional[T | str]`. For me the first one is more consistent.
6. Another question rises: `Optional[str]` vs `None | str` vs `str | None`. Here I'm not pretty sure, but the second options wins in my head by a small margin.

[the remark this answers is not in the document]
fscherf: Yeah! ``None | T | str`` makes a lot of sense to me! I agree, and i am looking forward to a PR :)

lona/events/view_event.py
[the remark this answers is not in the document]
maratori: Moreover you can omit quotes in type annotation using `from __future__ import annotations`, see [example](https://github.com/maratori/lona/commit/952b75e3d093e5c8316b9d197d0c23180be633f2).

tests/test_view_events.py
maratori: > but we cant just wait for the call to return

Yes, that's ok.

I meant to use `wait_for` instead of `sleep` + `assert`. Because actually you test that some state will be reached eventually.


One more comment. In my project it was convenient to use `wait_for` above. But here it seems better to create a context manager. Something like
```py
with eventually():
      assert ...
```

fscherf: Ah now i get it. I like that a lot! Could we add this to the lona pytest plugin? I think that would be helpful in test suites for lona applications as well

maratori: Yeah, why not.

maratori:
> I tried to reproduce this on multiple systems, but I had no luck.

I'm either. I tried [act](https://github.com/nektos/act), for example.
The good thing is that it reproduces steadily in github actions. Maybe [tmate](https://github.com/marketplace/actions/debugging-with-tmate) can help?

lona/events/view_event.py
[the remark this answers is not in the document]
maratori: Here it is #138

tests/test_view_events.py
fscherf: Ok. Can you create a PR for it? I would like to leave the sleep in there for now, to get this PR through. You would have a great show case for the new helper.

maratori: Sure

fscherf:
@maratori I added all of your feedback. tmate looks good, let's keep that in mind if pytest-timeout does not help with this kind of problems

themutt:
The APIs look nice :) Actually I'm not sure I can I'm confident enough to review this code so I will follow your discussion and test the PR.
My two cents:
Should there be some docs on this feature?
Also, do you think it would be useful to add an example in the `test_project` or in the docs?

fscherf:
@themutt no worries! The most important part is the API design, and if it makes sense, and you sure have an opinion on that.

Yes! This will feature will receive documentation when its thought,  for now, the test is the documentation :)

lona/app.py
```diff
@@ -236,7 +236,7 @@ def parse_command_line(self) -> dict:
 
         parser.add_argument(
             '--debug-mode',
-            choices=['messages', 'views', 'input-events'],
+            choices=['messages', 'views', 'input-events', 'view-events'],
```
maratori, line 239: Just for my understanding. Am I right that only one of this modes can be used in a moment? So I can't debug `messages` and `view-events`, for example.

lona/events/view_event.py
```diff
@@ -0,0 +1,15 @@
...
+
+if TYPE_CHECKING:
+    from lona.view import LonaView
+
+
+@dataclass
+class ViewEvent:
+    name: str
+    view_classes: list[type['LonaView']]
+    data: dict | None = None
```
maratori, line 15: nit: Maybe we can simplify it? `None` to `dict` conversion happens inside `fire_view_event` method.
```suggestion
    data: dict
```

lona/view_runtime_controller.py
```diff
@@ -315,3 +320,77 @@ def handle_lona_message(self, connection, window_id, view_runtime_id,
...
+
+        for view_runtime in self.iter_view_runtimes():
+            if view_classes and view_runtime.view_class not in view_classes:
+                continue
+
+            # if on_view_event is not defined by the view class
+            # it is unnecessary to start a thread
+            if view_runtime.view_class.on_view_event == LonaView.on_view_event:
+                view_events_logger.debug(
+                    '%r skipped (view class does not implement on_view_event()',
```
maratori, line 380: typo
```suggestion
                    '%r skipped (view class does not implement on_view_event())',
```

lona/view_runtime_controller.py
```diff
@@ -315,3 +320,77 @@ def handle_lona_message(self, connection, window_id, view_runtime_id,
...
+                    view_runtime.view,
+                )
+
+                continue
+
+            view_event = ViewEvent(
+                name=name,
+                data=data.copy(),
+                view_classes=view_classes,
```
maratori, line 389: Let's copy list as well. Just in case :)
```suggestion
                view_classes=view_classes.copy(),
```

maratori (commented):
@fscherf Thanks!

Last nitpicking from my side.

lona/app.py
fscherf: Thats right. If you wan't to see both you can use ``--loggers lona.messages lona.view_events``.

lona/events/view_event.py
fscherf: Good point. I will change the type to ``dict`` only.

themutt:
There is something that I didn't understand about`broadcast` in the tests. I see there is a `FireBroadcastEvent` that is sending `fire_view_event('broadcast', {'broadcast': True})`. What I don't understand: is there a specific argument for broadcast in the `fire_view_event()` method? How do the other views know that the broadcast should be intercepted?

maratori:
@themutt There are two `fire_view_event` methods: in `LonaView` and in `LonaServer`.

Method in server receives 3 arguments: `name`, `data` and target `view_classes`.
In the test the third argument is omitted. It means that event will be send to all existing views in the server.

Method is view receives only 2 arguments: `name` and `data`. And it always sends an event to `self.__class__` only.

fscherf:
@themutt, @maratori: one change in last minute: I will rename ``LonaView.fire_event()`` to ``LonaView.fire_view_event()`` to make naming consistent with ``Server.fire_view_event()``.

lona/view_runtime_controller.py
fscherf: Fixed

lona/view_runtime_controller.py
fscherf: fixed

fscherf:
@maratori i added all of your feedback and used ``lona.pytest.eventually`` instead of ``asyncio.sleep``

maratori (approved):
👍

fscherf:
@maratori, @themutt Thanks for reviewing! :) I will write  some documentation and then release 1.7.3

PR #138, opened 2021-10-01
types: use new type annotation syntax

maratori:
See PEP 563, 585 and 604.


fscherf:
@maratori Thanks for that, i like this syntax a lot!

PR #140, opened 2021-10-02
Add function `eventually()` for testing

maratori:
New function `eventually()` allows to wait for an expected state in tests. Iterator and context managers are used to be able repeating assertions.

Alternative approach is to accept function as an argument, and call it several times. But in this case user has to create a function inside test to make an assert. That's because `assert` is a statement and python doesn't allow it inside the `lambda`.

### DIFF
--- a/lona/pytest.py
+++ b/lona/pytest.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
-from asyncio import AbstractEventLoop
+from collections import AsyncIterator, Iterator, Callable
+from typing import AsyncContextManager, cast
+from asyncio import AbstractEventLoop, sleep
+from contextlib import asynccontextmanager
 from dataclasses import dataclass
-from collections import Callable
-from typing import cast
+import time
 
 from aiohttp.test_utils import TestClient
 from aiohttp.web import Application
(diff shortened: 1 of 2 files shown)


lona/pytest.py
```diff
@@ -6,31 +13,38 @@
...
-        lona_app = LonaApp(project_root or request.fspath)
+        lona_app = LonaApp(project_root or str(request.fspath))
         setup_app(lona_app)
 
         # setup client
-        def setup_aiohttp_app(loop):
+        def setup_aiohttp_app(loop: AbstractEventLoop) -> Application:
             lona_app.setup_server(loop=loop)
 
+            assert lona_app.aiohttp_app is not None
```
fscherf, line 43: Why is this necessary? This has nothing to todo with types also, so it should be mentioned in the commit message at least.

This assert can fail in end user test suites. Therefore it should have an assertion message

lona/pytest.py
```diff
@@ -73,6 +87,8 @@ def setup_aiohttp_app(loop):
 
         client = await aiohttp_client(setup_aiohttp_app)
 
+        assert server is not None
```
fscherf, line 93: Same here

lona/pytest.py
```diff
@@ -96,3 +99,56 @@ def setup_aiohttp_app(loop: AbstractEventLoop) -> Application:
...
+
+        @asynccontextmanager
+        async def attempt() -> AsyncIterator[None]:
+            nonlocal context_manager_was_used, success
+            context_manager_was_used = True
+
+            try:
+                yield  # execute assertions
+
+            except KeyboardInterrupt:
```
fscherf, line 139: Why are KeyboardInterrups excepted if they are reraised without further handling?

lona/pytest.py
```diff
@@ -96,3 +99,56 @@ def setup_aiohttp_app(loop: AbstractEventLoop) -> Application:
         )
 
     return setup_lona_project_context
+
+
+def eventually(timeout: float = 5, interval: float = 1) -> Iterator[AsyncContextManager]:  # NOQA: LN001
+    """
+        Wait for expected state in async test.
```
fscherf, line 106: The comment don't has to be indented this far

fscherf (changes requested):
Great PR! I am really looking forward to use this. Just some minor nitpicking

lona/pytest.py
maratori: Fixed

lona/pytest.py
maratori: I thought that it is necessary to be able interrupting tests. But It turned out that I was wrong. Removed this.

lona/pytest.py
maratori: > Why is this necessary?

That's a good question. Because from types perspective it is `Optional[Application]`, but should be not optional.
Before `lona_app.setup_server()` it is `None` and become real application after that.

I removed assertions and used `cast` instead.

fscherf:
@maratori Thanks for that! I will use this API in #137

PR #141, opened 2021-10-03
feat(tests): add lona_view_context pytest fixture

maratori:
New `lona_view_context` pytest fixture allows not to create separate function `setup_app` in tests.

@fscherf I'm not sure we need this. What do you think?

### DIFF
--- a/lona/pytest.py
+++ b/lona/pytest.py
@@ -54,6 +54,17 @@ def lona_app_context(request, aiohttp_client):
 
 
 @pytest.fixture()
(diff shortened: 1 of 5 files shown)


fscherf:
@maratori Interesting, but i am not sure if this is necessary. If i understand correctly, the only advantage of ``lona_view_context`` is that you don't have to add a route. I don't think it's worth new API, documentation etc

maratori:
Ok

PR #145, opened 2021-10-05
doc(demos): add view events demo

fscherf:


fscherf:
@MShekow Any thoughts?

doc/content/demos/multi-user-chat/multi-user-chat.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # choose a room
+        while True:
+            input_event = self.await_input_event(html=html)
+
+            # button
+            if input_event.node_has_id('new-room-name'):
+                new_room = input_event.node.value
+
+                if new_room in rooms:
```
MShekow, line 56: Couldn't this be simplified to a one-liner? `html.query_selector('#create-new-room').disabled = new_room in rooms`

fscherf: Yes that's right, but that's demo code, and i wanted it to be as verbose as possible

MShekow:
> @MShekow Any thoughts?

I like the demo :) - nothing comes to mind that I would change (I added a inline comment with a minor nit-pick).

doc/content/demos/multi-user-chat/multi-user-chat.py
fscherf: @MShekow Do you agree, or should i change it?

MShekow: The verbose version is a bit easier to understand, so you might as well leave it as it is ;)

doc/content/demos/multi-user-chat/multi-user-chat.py
```diff
@@ -0,0 +1,205 @@
...
+        while True:
+            self.await_click(html=html)
+
+            if not name.value:
+                continue
+
+            # check if name is already taken for this room
+            name_already_taken = False
+
+            for view in self.iter_objects():
```
maratori, line 140: nit: I think it's better to store user names in server state instead of iterating over the views.

fscherf: I absolutely agree, but usernames then would be taken indefinitely because  currently there is no good way to react to a user disconnecting from a view (i currently prepare a PR to fix that)

fscherf:
I created #156. Afterwards ``LonaView.iter_objects()`` is fully obsolete. I will come back to this PR after it is through.

fscherf:

PR #147, opened 2021-10-05
docs(html): add debouncing to input_delay description

MShekow:
See https://github.com/lona-web-org/lona/discussions/144


fscherf:
@MShekow Great! Thanks for that!

PR #151, opened 2021-10-05
fix(html): fix handling of generators in subnodes

fscherf:


lona/html/node.py
```diff
@@ -56,7 +57,7 @@ def __init__(self, *args, tag_name=None, self_closing_tag=None, **kwargs):
 
         # args (nodes)
         for arg in args:
-            if isinstance(arg, (list, tuple)):
+            if isinstance(arg, (list, tuple, GeneratorType)):
```
maratori, line 60: What about other iterable types?

fscherf: fixed

maratori (approved):

PR #153, opened 2021-10-06
html: fix handling of iterables

fscherf:
This patch fixes a bug i introduced with #151, and adds some tests


fscherf:

PR #161, opened 2021-10-11
feat(protocol): setup ping messages

fscherf:
Modern browsers like Chrome close websockets after a preconfigured timeout of
around five minutes of inactivity to save energy. This can lead to all sorts
of bad user experience, because all important state is part of the view in
Lona.
This patch adds ping and pong messages to the protocol and client
infrastructure to send pings in a configurable interval.


fscherf:
@maratori: Thanks for reviewing! :)

FTR: This has nothing to do with ``LonaView.ping()``. ``LonaView.ping()`` is meant to check if the server wants to shutdown

PR #164, opened 2021-10-12
views: fix HTTP pass through

fscherf:
HTTP pass through broke in e7e70dbd367016eb9b43b06ef258a14982780e48.
This PR fixes HTTP pass through again, and adds a test to avoid further regressions.


fscherf:

PR #167, opened 2021-10-19
html: fix multi selects

fscherf:
This pr fixes #165


fscherf:
@maratori great! Thanks for reviewing!

PR #173, opened 2021-10-29
pytest: fix race condition

fscherf:
This PR adds a test for ``LonaView.on_stop`` and ``LonaView.on_cleanup``, and fixes a lockup between the server and the application code.

@maratori This does not fix the pytest+playwright lockup problem entirely, but it seems to improve the situation. Previously the tests froze, which now seems fixed. There is still a teardown problem sometimes, but we make it through the tests.


maratori (approved):
👍

fscherf:

PR #178, opened 2021-11-09
prepare 1.8 release

fscherf:
This PR contains some small new features, bugfixes and removes some previously deprecated features


lona/view.py
```diff
@@ -45,24 +38,6 @@ def __init__(
         self._view_runtime: ViewRuntime = view_runtime
         self._request: Request = request
 
-    # objects #################################################################
-    @classmethod
-    def iter_objects(cls: type[V]) -> Iterator[V]:
-        # TODO: remove after 1.8
-
-        warnings.warn(
-            'LonaView.iter_objects() will be removed in 1.8',
```
maratori, line 54: Don't we need to remove this from docs as well?

lona/view_loader.py
```diff
@@ -66,15 +64,6 @@ def _run_checks(self, route: Route, view: type[LonaView]) -> None:
                     hook_name,
                 )
 
-        # TODO: remove after 1.8
-        if(isinstance(view, type) and
-           view.on_shutdown is not LonaView.on_shutdown):
-
-            warnings.warn(
-                'LonaView.on_shutdown() will be removed in 1.8',
```
maratori, line 74: The same as above

lona/view.py
fscherf: I will leave the docs for this method in the docs until 2.0, but i will add ``note: removed in 1.8`` like before. I do this usually when writing the changelog.

lona/view_loader.py
fscherf: The same as above :)

maratori (approved):
LGTM

laundmo:
LGTM, haven't tested but the code seems clean

laundmo (approved):

fscherf:
I found one last line using ``LonaView.on_shutdown`` while testing and fixed it.

@maratori, @laundmo: Thanks for reviewing!

mwx23:
Woohoo new `lona` version! I'm excited! :)

PR #182, opened 2021-11-16
fix(html): widget data: fix patch generating

fscherf:
The payload of patches have to be full copies of the original values to prevent
changes to the payload after the patch got created.

Previously some patches contained the original list or dict. When the dict or
list changed, the patch payload changed implicitly.


maratori (approved):
👍

fscherf:

PR #185, opened 2021-11-23
setup dependabot for testing

fscherf:
We had some problems with version ranges in test dependencies, so @maratori recommended using dependabot.
This PR splits all dependencies in dedicated files and and pins all versions.

@sobolevn: I enabled "Dependabot alerts" and "Dependabot security updates" in the project settings. Is there more i have to do to make dependabot work like expected?


sobolevn (commented):
Hi!

You can also add `.github/dependabot.yml` file. Like so: https://github.com/wemake-services/wemake-python-package/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/.github/dependabot.yml

You might need to add multiple entries with different `directory setting, like: tests/REQUIREMENTS.lint.txt`. One for each `REQUIREMENTS.*` file

Note that it also includes github-actions update.

fscherf:
@sobolevn Thanks for the example! I added a dependabot config. I assume the github-actions sections checks all configs in ``.github``. Is that correct?

sobolevn (approved):
Great! Thanks!

fscherf:
@sobolevn Great! Thanks for reviewing! :)

PR #188, opened 2021-12-13
fix import errors on windows

fscherf:
Currently Lona has different dependencies for production, packaging, linting and testing. All import errors during these stages get ignored by a try except block in the Lona top level module. This worked ok until now, but it covers up real import errors in production.
This PR adds all dependencies to all environments and removes the code that ignores all import errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install -U setuptools
-        pip install tox
+        pip install -r tests/REQUIREMENTS.lint.txt
 
     - name: Run Tox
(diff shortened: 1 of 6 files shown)


fscherf:
Hm! The linter works locally for me. The github action runs ``tox -e lint`` but tox seems not to reinstall all dependencies from ``tests/REQUIREMENTS.lint.txt``

@sobolevn do i have to remove some artifacts or reset the actions somehow?

themutt (approved):
I tested it on Windows 10 with python 3.10.1 and I confirm this is importing `LonaView` and `LonaApp` correctly

fscherf:
@themutt Great! Thanks for testing! :)

PR #189, opened 2022-01-06
add focus and blur events

fscherf:
This PR adds support for ``FOCUS`` and ``BLUR`` input events. It also contains a patch to fix current packaging problems.


maratori (approved):

fscherf:

PR #195, opened 2022-01-18
fix quoting in attribute dicts

fscherf:
Previously HTML representations looked like that:

    <input data-lona-node-id="1" type=&quot;checkbox&quot; />

This was because the whole attribute string was quoted and not every attribute
value individually.


maratori (approved):
👍

fscherf:
@maratori thanks for reviewing! :)

PR #204, opened 2022-01-26
improve testing

fscherf:
This PR adds some new tests, some new testing helper and documentation.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,3 +4,4 @@ source = lona
 
 omit =
     lona/__main__.py
+    lona/default_routes.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,6 @@
 [run]
 branch = True
 source = lona
+
+omit =
+    lona/__main__.py
--- a/doc/content/end-user-documentation/testing.rst
+++ b/doc/content/end-user-documentation/testing.rst
@@ -1,0 +1,221 @@
+
+
+Testing
+=======
+
+For testing Lona comes with a `pytest <http://docs.pytest.org>`_ plugin, based
+on pytest-aiohttp. The plugin contains two fixtures:
+`lona_app_context <#lona-app-context-example>`_ and
+`lona_project_context <#lona-project-context-example>`_.
+
+`lona_app_context <#lona-app-context-example>`_ spins up a test server using an
+API like the
+`Lona script interface </end-user-documentation/getting-started.html#running-lona-from-a-script>`_,
+`lona_project_context <#lona-project-context-example>`_ takes the project root
+and settings of a
+`Lona project </end-user-documentation/getting-started.html#starting-a-lona-project-from-the-project-template>`_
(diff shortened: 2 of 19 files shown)


doc/content/end-user-documentation/testing.rst
```diff
@@ -0,0 +1,176 @@
...
+`lona_project_context <#lona-project-context-example>`_.
+
+`lona_app_context <#lona-app-context-example>`_ spins up a test server using an
+API like the
+`Lona script interface </end-user-documentation/getting-started.html#running-lona-from-a-script>`_,
+`lona_project_context <#lona-project-context-example>`_ takes the project root
+and settings of a
+`Lona project </end-user-documentation/getting-started.html#starting-a-lona-project-from-the-project-template>`_
+to start the test server. Lona test servers run on an unprivileged random port.
+
```
SmithChart, line 18: In the following examples you use ``playwright`` as browser-interface. But you never introduce it here.  Maybe add it to the list of tools like ``pytest`` and ``pytest-aiohttp``.
Maybe add some words about ``playwright`` vs other browser-testing frameworks here?

fscherf: Good point! I added a section about browser testing and a synchronous code example for frameworks like selenium

doc/content/end-user-documentation/testing.rst
```diff
@@ -0,0 +1,221 @@
...
+The Lona testing infrastructure is meant to be used together with a browser
+testing library like `playwright <https://playwright.dev/python/docs/intro>`_
+or `selenium <https://selenium-python.readthedocs.io/>`_.  The test suite for
```
SmithChart, line 21: Nitpicking: Single double space here :-)

doc/content/end-user-documentation/testing.rst
SmithChart: Thanks! Looks good.

doc/content/end-user-documentation/testing.rst
fscherf: Fixed! Thanks!

.coveragerc
```diff
@@ -1,3 +1,10 @@
...
-source = lona
\ No newline at end of file
+source = lona
+
+omit =
+    lona/__main__.py
+    lona/default_routes.py
+    lona/logging.py
+    lona/command_line/*
+    lona/shell/*
```
laundmo, line 10: im not quite sure why these would absolutely need to be omitted from coverage. testing CLI-args and shel lcommands seems like something that we would want long-term.

fscherf: Hm, you are right. I removed everything but ``lona/default_routes.py`` and ``lona/__main__.py``

laundmo (approved):
LGTM

fscherf:
@SmithChart, @laundmo: Thanks for reviewing :)

PR #205, opened 2022-01-27
app: return the class from the route decorator

ukleinek:
Without the return in the decorator the effect is that after

  @app.route('/')
  class MyView(LonaView):
    ...

the name `MyView` is bound to `None`, because the decorator syntax is just
a nice way to write:

  class MyView(LonaView):
    ...
  MyView = app.route('/')(MyView)

With the return added, `MyView` holds the class as expected.


ukleinek:
force-push to (hopefully) make tox happy

fscherf:
@ukleinek: Great! Thanks for contributing! :)

PR #206, opened 2022-01-31
logo: fix bleeding background

rohieb:
Some SVG renderers have bugs when two areas are right next to each other, and render each area on their own with antialiasing in relation to the background color. This has the disadvantageous effect that the background color bleeds through to the foreground, which in our case is mostly visible on dark backgrounds.

To fix this, add a 1px outline with the same color as the fill color to each of those areas, so that neighboring areas effectively overlap with each other, thereby covering the background.

Compare screenshots in Firefox before:
0657a25a-828a-11ec-a686-507b9d024c54

and after:
f0ed8a5a-828a-11ec-b9b5-507b9d024c54

### DIFF
--- a/doc/content/logo.svg
+++ b/doc/content/logo.svg
@@ -12,8 +12,8 @@
    viewBox="0 0 200.00001 200"
    version="1.1"
    id="svg8"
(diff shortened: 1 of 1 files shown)


fscherf:
@rohieb Great! Thanks a lot for that! :)

PR #208, opened 2022-02-19
multiple fixes and better client debugging

fscherf:
This PR fixes page titles on daemonized views and adds a proper debug mode for the client


fscherf:
@maratori I had problems ``tests/test_static_files_widget.py`` failing again (this is the test we both debugged before because it failed randomly running in the CI). It seems it had something to do with multi-threading. I fixed it by implementing a hook using PEP 487 (https://www.python.org/dev/peps/pep-0487/)

lona/client_pre_compiler.py
```diff
@@ -48,33 +69,65 @@ def get_enums(self) -> Dict[str, Dict[str, Any]]:
...
+                    source_file,
+                )
 
-            template_context = {
-                'protocol': dumps(self.get_enums()),
-                'settings': dumps(self.get_settings()),
-            }
+                with open(path, 'w+') as f:
+                    f.write(file_content)
+                    f.close()
```
pandrey2003, line 113: This line of code may be redundant as the file will be closed by the context manager in `__exit__`

fscherf: Ack! :) fixed!

pandrey2003: Happy to have helped you :)
Just a fresh look

fscherf: Review is always welcome!

maratori (approved):
LGTM

fscherf:

PR #209, opened 2022-02-21
ci(pytest): cleanup tox.ini and pytest.ini

fscherf:
Previously the test root directory was configured both in the tox.ini and the
pytest.ini.
This patch moves the configuration into the pytest.ini and fixes some
whitespaces too.


maratori (approved):

fscherf:

PR #212, opened 2022-02-28
HTML: add server data

fscherf:
This PR adds ``node.server_data`` and contains some bugfixes and small features


doc/content/end-user-documentation/html.rst
```diff
@@ -382,6 +382,53 @@ and release it after writing.
...
+
+            for first_name, last_name, secret_id in DATA:
+                tr = Tr(
+                    Td(first_name),
+                    Td(last_name),
+                    events=[CLICK],
+                )
+
+                # set secret id
+                tr.user_data['secret_id'] = secret_id
```
maratori, line 419:
```suggestion
                tr.server_data['secret_id'] = secret_id
```

doc/content/end-user-documentation/html.rst
```diff
@@ -382,6 +382,53 @@ and release it after writing.
...
+                tr.handle_click = self.handle_click_on_tr
+
+            self.show(html)
+
+        def handle_click_on_tr(self, input_event):
+            # retrieve secret id
+            secret_id = input_event.node.user_data['secret_id']
```
maratori, line 426:
```suggestion
            secret_id = input_event.node.server_data['secret_id']
```

lona/html/abstract_node.py
```diff
@@ -102,6 +102,18 @@ def lock(self):
 
         return document.lock
 
+    # server data #############################################################
+    @property
+    def server_data(self):
+        if not hasattr(self, '_server_data'):
+            self._server_data = {}
```
maratori, line 109: What about concurrency? Don't we need to protect `server_data` with lock?

fscherf: Oh good point! This would be the only user accessible data structure that is not thread safe. 
We can reuse ``server.state`` for this. Should we call it ``node.state`` in this case?

doc/content/end-user-documentation/html.rst
fscherf: Ack! Fixed

doc/content/end-user-documentation/html.rst
fscherf: Ack! Fixed

lona/html/abstract_node.py
fscherf: I used ``lona.server_state.ServerState`` instead of an simple dict to make it thread-safe. In order to make the naming consistent i renamed it to ``node.state``

maratori (approved):
LGTM

fscherf:

PR #215, opened 2022-03-02
feat(imports): add support for top-level imports

fscherf:
Previously top-level import strings like 'json' were not possible.


maratori (approved):

fscherf:

PR #216, opened 2022-03-04
server: Allow loading of state in settings

SmithChart:
Until now there was no way to fill ``server.state`` before the first actual view has been initialized.

With this changes it is now possible to define or load the initial values for ``server.state`` in the ``settings.py``.

In my case I have a rather simple web application that keeps track of a queue of images.
One view shows this queue of images ("frontend") and another view allows to configure this queue ("backend").
Both views share the current queue using ``server.state``. But: I did not find an elegant way to load my initial state:
I could do it in ``handle_request`` but this would mean code duplication between the views.

For me the best way to handle this is to allow the user to define the *initial_state* of the ``server.state``.
And since I can run arbitrary python-magic in ``settings.py``this seems a perfect fit.

### DIFF
--- a/doc/content/end-user-documentation/settings.rst
+++ b/doc/content/end-user-documentation/settings.rst
@@ -170,6 +170,13 @@ Views
     :name: FRONTEND_VIEW
     :path: lona.default_settings.FRONTEND_VIEW
 
+.. setting::
+    :name: INITIAL_SERVER_STATE
+    :path: lona.default_settings.INITIAL_SERVER_STATE
+
+    This dict, if present, is copied to ``server.state``.
+    This is a convenient way to define or load server state on startup without
+    the need to write a specific middleware.
 
 Error Views
 -----------
--- a/lona/default_settings.py
+++ b/lona/default_settings.py
@@ -57,6 +57,7 @@ SESSIONS_KEY_RANDOM_LENGTH = 28
 # views
 CORE_FRONTEND_VIEW = 'lona.default_views.FrontendView'
 FRONTEND_VIEW = ''
+INITIAL_SERVER_STATE: dict = {}
 
 # error views
 CORE_ERROR_403_VIEW = 'lona.default_views.Error403View'
--- a/lona/server.py
+++ b/lona/server.py
@@ -4,6 +4,7 @@ from typing import overload, Callable, TypeVar, cast, Any
 from concurrent.futures import Future
 from collections.abc import Awaitable
 from functools import reduce
+from copy import deepcopy
 import contextlib
 import operator
 import logging
@@ -102,8 +103,9 @@ class LonaServer:
 
         # setup state
         server_logger.debug('setup state')
-
-        self._state = State(initial_data={})
+        self._state = State(
+            initial_data=deepcopy(self.settings.INITIAL_SERVER_STATE),
+        )
 
         # setup routing
         server_logger.debug('setup routing')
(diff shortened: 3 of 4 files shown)


fscherf:
Hi @SmithChart,
the standard way to achieve this is to write a middleware and setup ``server.state`` in a ``on_startup`` hook, but I like your idea. The patches look ok to me, just some minor feedback.

lona/server.py
```diff
@@ -102,8 +102,9 @@ def __init__(self, app, project_root, settings_paths=None,
 
         # setup server state
         server_logger.debug('setup server state')
-
-        self._state = ServerState(initial_data={})
+        self._state = ServerState(
+            initial_data=self.settings.INITIAL_STATE,
```
fscherf, line 106: We cant just pass the original settings variable in here because this way you could change ``server.state`` using ``server.settings.INITIAL_STATE`` and bypass all locks. Please use deepcopy.

lona/default_settings.py
```diff
@@ -56,6 +56,7 @@
 # views
 CORE_FRONTEND_VIEW = 'lona.default_views.FrontendView'
 FRONTEND_VIEW = ''
+INITIAL_STATE = {}
```
fscherf, line 59: This needs a type annotation to make the linter happy again

doc/content/end-user-documentation/settings.rst
```diff
@@ -170,6 +170,10 @@ Views
     :name: FRONTEND_VIEW
     :path: lona.default_settings.FRONTEND_VIEW
 
+.. setting::
```
fscherf, line 173: Could you please add a short sentence of documentation here?

fscherf (changes requested):

lona/server.py
SmithChart: Fixed.
And also followed your refactoring from ``ServerState`` to ``State``.

doc/content/end-user-documentation/settings.rst
SmithChart: Done

SmithChart:
@fscherf Well... Now we started editing at the same time.

It seems my patch still fails CI. So take what ever you like more.

lona/default_settings.py
SmithChart: Fixed

doc/content/end-user-documentation/settings.rst
```diff
@@ -170,6 +170,13 @@ Views
...
+    :name: INITIAL_STATE
+    :path: lona.default_settings.INITIAL_STATE
+
+    This dict, if presnet, is copied to ``server.state``.
```
fscherf, line 177: s/presnet/present/

lona/default_settings.py
```diff
@@ -57,6 +57,7 @@
 # views
 CORE_FRONTEND_VIEW = 'lona.default_views.FrontendView'
 FRONTEND_VIEW = ''
+INITIAL_STATE: Dict[Any, Any] = {}
```
fscherf, line 60: Just use ``dict`` here. Can you rename ``INITIAL_STATE`` to ``INITIAL_SERVER_STATE``?

fscherf (changes requested):

fscherf:
@SmithChart: Do you want to add my change requests to this pr? If not i will merge my branch with you updated patch (i am fine with both, they do pretty much the same)

lona/default_settings.py
SmithChart: Updated. Please take another look at the type annotations :-)

SmithChart:
> @SmithChart: Do you want to add my change requests to this pr? If not i will merge my branch with you updated patch (i am fine with both, they do pretty much the same)

Sorry. Totally overlooked the activity here. Thanks for the bump!

fscherf (approved):

fscherf:
@SmithChart: No Problem. Thanks for contributing :)

PR #218, opened 2022-03-12
fix(HTML): Document: fix incorrect usage of Document.apply

fscherf:
This patch fixes a bug, introduced in

    ff7a0ab8316c ("fix(views): fix page titles on daemonized views")

ff7a0ab8316c introduced a new, optional argument to Document.apply, named
'title' and broke all code that calls Document.apply without named arguments.


maratori (approved):

fscherf:

PR #223, opened 2022-05-22
feat(server/routing): Add support for trailing wildcards

SmithChart:
This change adds support for wildcard routes.
Wildcards can only be used at the end of a route.

A wildcard can be used to Route all URLs beginning with a
given path to a view or a Django application.

Consider the following example for the Django Admin:
The easy way to route the Django Admin is to have a fallback route at
the end of the routing table.
But this is not possible if there is another route that matches the
URL.

In this case a wildcard match can be used:

```
routes = [
    # (...)

    # More specific match
    Route('/admin(*)', wsgi_handler, http_pass_through=True),

    # Broad match
    Route('/<output>(/)', 'views/home.py::HomeView'),
]
```

### DIFF
--- a/doc/content/end-user-documentation/views.rst
+++ b/doc/content/end-user-documentation/views.rst
@@ -157,6 +157,22 @@ Trailing Slashes
     ]
 
 
+Wildcards
+~~~~~~~~~
+
+Wildcards allow to glob the end of a route.
+
+.. code-block:: python
+
+    # routes.py
+
+    from lona import Route
+
+    routes = [
+        Route('/foo(*)', 'views/my_view.py::MyView'),
+    ]
+
+
 Request Objects
 ---------------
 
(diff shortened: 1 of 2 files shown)


fscherf:
@SmithChart

Can't the same effect be achieved using a route like this:
```
Route('/admin<path_info:.*>', wsgi_handler, http_pass_through=True)
```
?

SmithChart:
That works. But: is that intended behavior?

If yes: works for me. But I'll come back with a documentation PR :-D

fscherf:
@SmithChart: Absolutely! :) Lona routes consist of named regex patterns like `<number:[0-9]+>`. The default pattern is `[^/]+` (any character but `/`, because URLs are divided by slashes most of the time). The pattern can be any valid regex to make things like this work.

That the name has to be `path_info` in this case is a implemention detail of `aiohttp-wsgi` (at least it was, last time o checked. Maybe it's not required anymore).

SmithChart:
Well, in this case I'll come forward with an update for the documentation. Thanks for the clarification.

PR #224, opened 2022-05-26
WIP: feat(LonaServer): Allow to set aiohttp client_max_size

### LINKED ISSUE #222: Correct way to set aiohttp's client_max_size ?
I am working on an application where I want to upload files using POST. But files can exceed aiohttp's default client_max_size (https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.Application ) limit.

I am running my server from the command line (aka the Makefile). So the right place to set this limit would be here: https://github.com/lona-web-org/lona/blob/master/lona/command_line/run_server.py#L29

What do you think should be the correct way to set this limit? Add another command line option? Any opinions?

SmithChart:
With this change it is now possible to set the aiohttp
``client_max_size`` during the creation of the ``Application``. This
value controls the maximum size of a POST request in bytes.

This change basically moves the creation of the aiohttp ``Application``
from ``run_server.py`` and ``app.py`` into the ``LonaServer``.

TODO:
* In ``app.py`` and ``pytest.py`` the ``loop`` was passed into the
  Application. This does not happen with my implementation.
  In my understanding aiohttp should grab the same loop as we would
  provide.
  But this needs a looks from some async-expert.
* ``run_server.py`` and ``app.py`` both need access to the
  ``Application``.
  I am currently accessing the private member ``server._app`` to
  archive this - as is already done with ``server._loop`` and
  ``server._worker_pool``.
  Pro: We are hiding an internal from _normal_ users.
  Con: Accessing private members from outside looks wrong.
* Documentation

This would close #222 

Let me know what you think.


SmithChart:
> @SmithChart: looks good to me. There seems to be no downside to create the app directly in the LonaServer constructor. Where does the default for `AIOHTTP_CLIENT_MAX_SIZE` come from? 

That is the AIOHTTP default value for this parameter.

> Can you add it to the documentation?

Ack. I will update my PR with documentation.

SmithChart:
> > Can you add it to the documentation?
> 
> Ack. I will update my PR with documentation.

Documentation and commit message updated. Please Review.

lona/app.py
```diff
@@ -462,6 +461,5 @@ def run(
         # start server
         run_server(
             args=server_args,
-            app=self.aiohttp_app,
```
fscherf, line 465: Why is this line removed?

doc/content/end-user-documentation/settings.rst
```diff
@@ -224,3 +224,21 @@ Testing
...
     :path: lona.default_settings.TEST_INPUT_EVENT_TIMEOUT
+
+Server
+------
+
+.. setting::
+    :name: AIOHTTP_CLIENT_MAX_SIZE
+    :path: lona.default_settings.AIOHTTP_CLIENT_MAX_SIZE
+
+    .. note::
```
fscherf, line 235: Can you please remove this message? The next version could be 1.11 as well.

fscherf (changes requested):
Looks good to me! Only some minor nitpicks. Besides them this looks ready to merge.

Can you remove the prefix "WIP: " when you are done?

fscherf:
@SmithChart: Whats the state of this PR?

doc/content/end-user-documentation/settings.rst
SmithChart: I hope I have understood correctly. I have now  removed the ``..note::``section.
I have updated my branch accordingly.

lona/app.py
SmithChart: The ``app`` is now created inside the ``LonaServer`` (see https://github.com/lona-web-org/lona/pull/224/files#diff-71b8e0eb0eada6eab6aa80a7d9faa08f0a543d3b9c8fc4e1c425f34b9d6f7b51R100 ). This way we have access to the settings and know the value of ``AIOHTTP_CLIENT_MAX_SIZE``.

This argument to ``command_line/run_server.py::run_server()`` has also been removed (see https://github.com/lona-web-org/lona/pull/224/files#diff-5096a6e3173b4dcd96c074d6a72b4cbf0c76af7b6581ef36144abe35dc06c4fdR22 ).
To still have access to the aiohttp-server app I just grab it from the internal state of the ``LonaServer`` (see here https://github.com/lona-web-org/lona/pull/224/files#diff-5096a6e3173b4dcd96c074d6a72b4cbf0c76af7b6581ef36144abe35dc06c4fdR35 ). This is not really elegant but is in line with the code around it.

SmithChart:
Sorry for weeks of silence. First some intense weeks at work and afterwards some days vacation.
*WIP*-flag is removed. And I have also answered your nitpicks.

fscherf:
@SmithChart: No worries :) I just wanted to make sure you are not waiting for me

PR #225, opened 2022-05-31
fix(shell): fix incompatibility with rlpython 0.10

fscherf:
rlpython version prior to 0.10 had separate globals and locals. 0.10 merged
them into a global namespace named globals.

This patch changes all code that reads or writes repl.locals.


maratori (approved):

fscherf:

PR #226, opened 2022-06-09
fix(doc): Show that custom URL-pattern can match any character

SmithChart:
In https://github.com/lona-web-org/lona/pull/223 I showed that I did not
understand that a custom pattern can be used to create a wildcard match.

This change updates the documentation to contain some more words about
custom patterns.

Signed-off-by: Chris Fiege <chris@tinyhost.de>


fscherf:
Great! Thanks for contributing :)

PR #233, opened 2022-09-06
feat(TemplatingEngine): Allow to set-up jinja environment

SmithChart:
This change allows a user to create an own TemplatingEngine with a
custom set-up Jinja2-Environment.

The lona default TemplatingEnine uses Jinja2 with default paramenters.
This is fine if uses inside lona to render HTML for the client.

For applications where templates are rendered for other languages, e.g.
Latex, the default settings may be in the way.
With this change a user can create their own instance of TemplatingEninge
with specific settings, e.g. overriding ``block_start_string`` and still
have the filters and paths available as set up in the settings.

Useage example:
```
engine = lona.templating.TemplatingEngine(
    server,
    jinja_environment_kwargs={
        'block_start_string': r'<%',
        'block_end_string': r'%>',
        'variable_start_string': r'<<',
        'variable_end_string': r'>>',
    },
)

content = engine.render_template(
    template_name,
)
```

### DIFF
--- a/lona/templating.py
+++ b/lona/templating.py
@@ -76,8 +76,11 @@ class Namespace:
 
 
 class TemplatingEngine:
-    def __init__(self, server):
+    def __init__(self, server, jinja_environment_kwargs=None):
         self.server = server
+
+        if not jinja_environment_kwargs:
+            jinja_environment_kwargs = {}
 
         self.template_dirs = (self.server.settings.TEMPLATE_DIRS +
                               self.server.settings.CORE_TEMPLATE_DIRS)
@@ -96,6 +99,7 @@ class TemplatingEngine:
                 self.template_dirs,
                 followlinks=True,
             ),
+            **jinja_environment_kwargs,
         )
 
         # load extra filters


fscherf:
@SmithChart: sorry that it took me so long, i am currently on vacation.

Interesting! What are your usecases for such a feature? Do you want to reset `server.templating_engine`?

SmithChart:
I do not want to reset the ``server.templating_engine``.
I want to be able to have a second templating_eninge with a different configuration:

The lona-default templating_engine is fine for HTML and plain text.

But for templating LaTeX another Jinja2 configuration would be better. For my LaTeX templates I replace the ``{{`` with ``<<``. This way I don't have collisions with the ``{`` already used by LaTeX.
My LaTeX-Wrapper than has a local ``templating_engine`` with a different Jinja2 configuration - but otherwise behaves in the same way as the ``server.templating_eninge`` does.

SmithChart:
Any opinions on this?

fscherf:
@SmithChart: I am not sure if i want this. The templating engine class is a not documented, internal class, that was never intended to be used in application code, or to be instantiated more than once. Also the code interface was planed to be independent from Jijna2, so i can swap Jinja2 with another templating engine in the future if needed.

SmithChart:
The documentation already states that there is a jinja2 in there somewhere. See https://lona-web.org/end-user-documentation/settings.html#templating . 

Having ``TEMPLATE_EXTRA_CONTEXT``  and  ``TEMPLATE_EXTRA_FILTERS`` would imply that a future jinj2-replace would have to be somewhat of a drop-in replacement anyway.

That said: I am not sure if replacing jinja2 would go without breaking any APIs anyway :-)

fscherf:
I thought about this request and i don't want this feature. I don't want to support using internal classes in end-user code. If you wan't to do it anyway, you can subclass the templating engine in your project.

PR #237, opened 2022-10-06
Fix `dependabot.yml` to work with dev deps

sobolevn:
As you can see, your dependecies are not updated.
Dependabot expects relative paths, not absolute ones.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,14 +2,14 @@ version: 2
 
 updates:
 - package-ecosystem: pip
-  directory: "./tests/REQUIREMENTS.lint.txt"
+  directory: "./tests/"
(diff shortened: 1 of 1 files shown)


fscherf:
@sobolevn: Great! Thanks :)

fscherf:
@sobolevn: dependabot opened 22 pull-requests since this pull-request is merged. Is this wanted behavior?

sobolevn:
Yes, because we have a lot of outdated deps 🙂

fscherf:
Fair enough :D 
Do you have any recommendations on how to merge the prs, fix those which break the test suite and which order to choose? 
For example: can i just merge all prs where the checks are green?

PR #260, opened 2022-10-09
html: make widgets obsolete

fscherf:
Like discussed in https://github.com/lona-web-org/lona/discussions/228, this PR adds all features of `lona.html.Widget` into `lona.html.Node`, to make widgets obsolete, so they can be removed in Lona 2

### DIFF
--- a/doc/content/end-user-documentation/html.rst
+++ b/doc/content/end-user-documentation/html.rst
@@ -819,12 +819,13 @@ handle the event.
 Frontend Widgets
 ~~~~~~~~~~~~~~~~
 
-Widgets can define a Javascript based frontend widget, to include client side
-code. This is useful to integrate with third party Javascript libraries.
+Widgets and nodes can define a Javascript based frontend widget, to include
+client side code. This is useful to integrate with third party Javascript
+libraries.
 
 To communicate between the backend widget and the frontend widget, the backend
-can set its state in ``Widget.state``, a dict like object, and the frontend
-can issue events with custom data.
+can set its state in ``Widget.state``, or in ``Node.widget_data`` a dict like
+object, and the frontend can issue events with custom data.
 
 .. code-block:: python
 
@@ -847,6 +848,31 @@ can issue events with custom data.
             ]
 
             self.data = {'foo': 'bar'}
+
+
+.. code-block:: python
+
+    # my_node.py
+
+    from lona.static_files import Script
+    from lona.html import Div
+
+    class MyNode(Div):
+        WIDGET = 'MyFrontendWidget'
+
+        STATIC_FILES = [
+            # the path is always relative to the current file
+            Script(name='MyFrontendWidget', path='my_frontend_widget.js'),
+        ]
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+
+            self.nodes = [
+                Div('foo'),
+            ]
+
+            self.widget_data = {'foo': 'bar'}
 
 
 .. code-block:: javascript
--- a/lona/client/_lona/dom-renderer.js
+++ b/lona/client/_lona/dom-renderer.js
@@ -48,6 +48,8 @@ export class LonaDomRenderer {
             var node_style = node_spec[5];
             var node_attributes = node_spec[6];
             var node_child_nodes = node_spec[7];
+            var widget_class_name = node_spec[8];
+            var widget_data = node_spec[9];
 
             var node = document.createElement(node_tag_name);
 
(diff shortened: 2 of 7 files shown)


test_project/views/frontend/widget_data.py
```diff
@@ -1,17 +1,19 @@
...
+class TestWidget(Div):
     STATIC_FILES = [
         Script(name='TestWidget', path='widget_data.js'),
     ]
 
-    FRONTEND_WIDGET_CLASS = 'widget_data'
+    WIDGET = 'widget_data'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
SmithChart, line 15: It's a shame that we have to introduce some new boilerplate here. Maybe that's just a bit for the documentation :)

test_project/views/frontend/widget_data.py
```diff
@@ -20,7 +22,7 @@ def __init__(self):
         ]
 
     def update_state(self):
-        self.server_state.set_text(dumps(self.data))
+        self.server_state.set_text(dumps(self.widget_data))
```
SmithChart, line 25: I am not sure if this test view is running correctly. From what the view tells me the ``client-state``-node should always contain the same value as the server side?
I don't see that behavior here. The  ``client-state ``-div is always empty.

test_project/views/frontend/widget_data.py
```diff
@@ -60,54 +62,54 @@ def handle_request(self, request):
 
         while True:
             # list
-            widget.data = {'list': []}
+            widget.widget_data = {'list': []}
```
SmithChart, line 65: Just nitpicking:
It would probably be better style to already set this value in ``TestWidget.__init__()``. This way it will always be available when calling ``TestWidget.update_state()``.

(But this is not relevant to the test case here.)

SmithChart (commented):
Thanks for implementing! Overall this looks really good. 

Let me try to summarize:
* Every HTML-Node can no be a Widget by setting ``self.nodes`` to a list of nodes, right? And the Node itself is the root-node of that widget.
* `` Node.WIDGET`` defines the identifier for the frontend widget javascript?

I'll give this a spin with my real-world widgets later on.

test_project/views/frontend/widget_data.py
fscherf: Yes, i know exactly what you mean. I had multiple prototypes with something like a `get_nodes()`` method or something, but to be a full widget replacement we have to have support for custom constructors. And to call `super()` in your constructor is pythonic and unsurprising at least.
I am open to proposals though.

fscherf:
> * Every HTML-Node can no be a Widget by setting `self.nodes` to a list of nodes, right? And the Node itself is the root-node of that widget.

Yes.

> * ` Node.WIDGET` defines the identifier for the frontend widget javascript?

Yes. I was never truly happy with this wording anyway. We have nodes, which describe "simple" nodes, we have widgets that describe collections of nodes, but behave similar to nodes and we have widgets in the frontend which are javascript classes. In the next version, there should be only nodes on the server side, and the word "Widget" describes only the javascript classes on the frontend side.

test_project/views/frontend/widget_data.py
fscherf: Ack! Fixed

test_project/views/frontend/widget_data.py
fscherf: Ack! I added this view to my refactoring list, but i don't want to change it to much in this pr.

fscherf:
@SmithChart: could you have a second look?

fscherf:
I though a lot about how to release this, and i don't like that this PR would mean breaking changes. I think we should only merge the first patch, because it's an additional change, add documentation and a deprecation warning the the "old" widget API. The port of `html.HTML` from a widget to a node should be part of Lona 2.

@SmithChart, @maratori, @laundmo, @korantu

maratori (approved):
I like this unification.

The only question from my side. Have you decided not to deprecate old widgets?

maratori:
@fscherf as far as I understood, you want to remove `Widget` class in lona 2.
What are you going to do with `HTML`?

Now it simply is a list of nodes (without root node). Will we lose such feature and have to deal with root node in all cases?

fscherf:
@maratori: I am not sure about how to make smooth transition. This patch adds functionality to make widgets obsolete. I have a patch that changes `lona.html.HTML` to be a function that returns only one node (because there will be no widgets in the future, there has to be exactly one root node everytime (if `lona.html.HTML` parses more than one, it wraps all nodes into one `div`))
That would be a breaking change, therfore i think it should be part of lona 2

fscherf:
Widgets are virtual nodes pretty much. The problem is they make the rendering in the Javascript client (at least in the current implementation) much more complex. Thats pretty much what i want to get rid of

fscherf:
> I'm not sure `div` is fine in all cases.

Do you have an example? I know it limits the application code to say "everything has to have exactly one root node, and if you don't define one, i wrap your nodes into one div" but other component frameworks seem to do just that an doesn't seem to be an issue.

If we decide we need something like widgets or virtual nodes that are collections of nodes, the patch in this pr is obsolete

maratori:
> Do you have an example?

No. Let's merge this PR.

PR #276, opened 2022-10-26
html: select: refactor API

fscherf:
This PR refactors the `lona.html.Select` API to be more concise and convenient.

### DIFF
--- a/doc/content/end-user-documentation/html.rst
+++ b/doc/content/end-user-documentation/html.rst
@@ -633,40 +633,66 @@ Select
 
 .. code-block:: python
 
-    from lona.html import Select
+    from lona.html import Select, Option
 
-    Select(
-        values=[
-            # value, label, is_selected
-            ('foo', 'Foo', True),
-            ('bar', 'Bar', False),
-        ],
+    select = Select(
+        Option('Option 1', value='1'),
+        Option('Option 2', value=2),
+        Option('Option 3', value=3.0, selected=True),
     )
 
-**Init Arguments:**
+    # disable first option
+    select.options[0].disabled = True
+
+    # select second option by selected property
+    select.options[1].selected = True
+
+    # select second option by value property
+    select.value = 2
+
+A ``Select`` consist of one or more ``Option`` objects, which hold information
+on value, selection state and disabled state.
+
+``Option`` objects consist of a label text and a value. The value can be
+anything. If ``Option.render_value`` is set, which is set by default, the
+content of ``Option.value`` gets typecasted to a string and rendered into the
+HTML tree. This can be disabled if the actually values of the select shouldn't
+be disclosed to end users.
+
+``Select.value`` returns the value of the option that is currently
+selected. If the ``Select`` is a multi select, ``Select.value`` returns a
+tuple of all selected options values.
+
+An option can be selected by setting ``Select.value`` to the value of the
+option that should be selected, or by setting ``Option.selected``. If the
+select is no multi select, all other options get unselected automatically.
+
+**Select Attributes:**
 
 .. table::
 
-    ^Name             ^Default Value      ^Description
-    |values           |None               |(List of Tuples) Initial values
-    |bubble_up        |False              |(Bool) Pass input events further
-    |disabled         |False              |(Bool) sets the HTML attribute "disabled"
-    |multiple         |False              |(Bool) Enables multi selection
-    |*args            |()                 |Node args
-    |**kwargs         |{}                 |Node kwargs
+    ^Name              ^Description
+    |value             |Value of the currently selected option or tuple of values of selected options
+    |values            |(Tuple) tuple of all possible values
+    |options           |(Tuple) tuple of all options
+    |selected_options  |(Tuple) tuple of all selected options
+    |disabled          |(Bool) sets the HTML attribute "disabled"
+    |multiple          |(Bool) Enables multi selection
+    |id_list           |(List) contains all ids
+    |class_list        |(List) contains all classes
+    |style             |(Dict) contains all styling attributes
 
-**Attributes:**
+**Option Attributes:**
 
 .. table::
 
-    ^Name       ^Description
-    |values     |(List of Tuples) All options
-    |value      |Currently set value
-    |disabled   |(Bool) sets the HTML attribute "disabled"
-    |multiple   |(Bool) Enables multi selection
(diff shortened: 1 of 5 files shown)


doc/content/end-user-documentation/html.rst
```diff
@@ -633,40 +633,66 @@ Select
...
+    select.options[1].selected = True
+
+    # select second option by value property
+    select.value = 2
+
+A ``Select`` consist of one or more ``Option`` objects, which hold information
+on value, selection state and disabled state.
+
+``Option`` objects consist of a label text and a value. The value can be
+anything. If ``Option.render_value`` is set, which is set by default, the
```
maratori, line 657: Don't see `render_value` in code

lona/html/data_binding/select.py
```diff
@@ -80,73 +154,109 @@ def readonly(self, new_value):
...
+            values = ()
 
-                return value
+            for option in self.selected_options:
+                values += (option.value, )
 
             if not self.multiple:
-                value = value.pop()
+                if not values:
+                    return self.options[0].value
```
maratori, line 211: There may be no options at all

lona/html/data_binding/select.py
```diff
@@ -80,73 +154,109 @@ def readonly(self, new_value):
...
 
     @value.setter
     def value(self, new_value):
-        if not isinstance(new_value, list):
-            new_value = [new_value]
+        with self.lock:
+            if not isinstance(new_value, list):
+                new_value = [new_value]
+
+            old_values = self.values
```
maratori, line 223: nit:
```suggestion
            values = self.values
```

lona/html/data_binding/select.py
```diff
@@ -80,73 +154,109 @@ def readonly(self, new_value):
...
+        with self.lock:
+            option._select = self
+            self.nodes.append(option)
+
+    def remove_option(self, option):
+        with self.lock:
+            option._select = None
+            self.nodes.remove(option)
+
+    def select_all(self):
```
maratori, line 254: Maybe need to throw exception for non-multi select?

lona/html/data_binding/select.py
```diff
@@ -80,73 +154,109 @@ def readonly(self, new_value):
...
-    # value
+    # selected options
     @property
-    def value(self):
+    def selected_options(self):
         with self.lock:
-            value = []
+            options = ()
+
+            for option in self.options:
```
maratori, line 189: Let's optimize a bit: `self.options` may be calculated only once in this method.

lona/html/data_binding/select.py
```diff
@@ -80,73 +154,109 @@ def readonly(self, new_value):
...
+                new_value = [new_value]
+
+            old_values = self.values
+
+            for value in new_value:
+                if value not in old_values:
+                    raise RuntimeError(f'unknown value: {value}')
 
+            for option in self.options:
+                option._set_selected(option.value in new_value)
```
maratori, line 230: What if there are several options with the same value? I think we should select only the first one (as in browser).

lona/html/data_binding/select.py
fscherf: Ack, Fixed!

lona/html/data_binding/select.py
fscherf: Ack, Fixed!

lona/html/data_binding/select.py
fscherf: Ack, Fixed!

lona/html/data_binding/select.py
fscherf: Yes i agree. Like in your test

lona/html/parsing.py
```diff
@@ -143,7 +143,7 @@ def handle_starttag(self, tag, attrs, self_closing=False):
         node_kwargs['self_closing_tag'] = self_closing
 
         # setup node
-        for key in ('id', 'class', 'style'):
+        for key in ('id', 'class', 'style', 'value'):
```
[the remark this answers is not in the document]
fscherf: @maratori: That means:

```python
select = Select(
    Option('Foo', value='foo', selected=True),
    multiple=False,
)

option = Option('Bar', value='bar', selected=True)

select.add_option(option)

assert option.selected == False
```

Right?

I agree that this behavior would be consistent with the API i proposed, but maybe that's a little bit to much magic because  what should we do in cases like these:

```python
select = Select(
    Option('Foo', value='foo', selected=True),
    Option('Bar', value='bar', selected=True),
)

select.multiple = True

assert select.options[0].selected == True
assert select.options[1].selected == True  # fails
```

In the browser that would work. Do we want this behavior? Or do we want `Option.selected` to be dumb and the "clever" metrics like "return the first selected option if multiple are selected and it's no multi-select" are in `Select.value`, `Select.selected_values` ?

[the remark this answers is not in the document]
maratori: > Right?

This seems to be not related to my original comment.

To answer your great question, I need to take some time to play around with this API and with browser.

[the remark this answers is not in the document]
fscherf: > This seems to be not related to my original comment.

No? If i read your example correctly, you set two options to selected in the HTML string but assert only one `selected == True` (b and c). Am i missing something?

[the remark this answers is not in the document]
maratori: I haven't run test, but from reading the code I think parsing string HTML doesn't work correctly with new select API. That's because `Select.append` behave differently from `Select.add_option`.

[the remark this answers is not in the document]
fscherf: Oh! You are right. Maybe a second reason to strip the proposed `Option.selected` API

fscherf:
RFC: I added a patch that removes the vast majority of magic i added before. The more i think about this i think less is more here, and it is least surprising to just emulate what the browser does here.

`Option.selected` just sets or unsets `selected` in `Option.attributes` now and does not interact with the surrounding `Select`. If you decide to use a non-multiple select, but select multiple options, you can do that, but `Select.value` will return the value of the first selected option.

@maratori, @SmithChart, @grigolet: Which approach do you like better?

fscherf:
I created #329, on top of `master`, that contains all feedback up to this point, and using my second implementation as base. This way we make the new implementation usable before Lona 2, making the transition easier.

PR #286, opened 2022-11-06
remove widget API

fscherf:
This PR removes the widget API, refactors the JavaScript rendering code and adds a browser-based rendering test.

The code and the commits are work in progress, but i would appreciate some early comments.

To test the rendering, i added a rendering test view to the test project (`/frontend/rendering`), that uses every rendering feature Lona offers, step by step. The original node tree get's stored in `server.state['rendering-root']`. The test in the test-suite runs this view and compares the node tree in `server.state` against the rendered HTML using playwright. To make this work i added a patch that makes Lona nodes comparable like this: `Div() == Div(); Div() != Div(a=1)`.
One problem with this approach is that `Node.style` can't be tested like this, because the value of the style property gets stored outside the DOM tree in the browser. I am not sure how to solve this.

@maratori, @grigolet, @SmithChart, @sobolevn


fscherf:
The patches are now ready for reviewing. I removed the Widget API, added a rendering test and did much refactoring on the JavaScript rendering code.

lona/html/abstract_node.py
```diff
@@ -45,6 +45,16 @@ def get_all_node_classes():
             *AbstractNode._subclasses,
         ]
 
+    # comparisons #############################################################
+    def __eq__(self, other):
```
maratori, line 49: Is this really necessary?

tests/test_rendering.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # start rendering test view
+        await page.goto(context.make_url('/frontend/rendering/'))
+        await page.wait_for_selector('#lona>div>h2:has-text("Rendering Test")')
+
+        rendering_root_element = page.locator('#lona #rendering-root')
+
+        # get total rendering steps
+        total_steps_element = page.locator('#lona #step-label #total')
+        total_steps = int(await total_steps_element.inner_text())
```
maratori, line 28: It's better to use expected value instead of relying to UI. For example test won't fail if there will be `0` in the page.
```suggestion
        total_steps = 21
```

tests/test_rendering.py
```diff
@@ -0,0 +1,42 @@
...
+
+        for step in range(1, total_steps):
```
maratori, line 30:
```suggestion
        for step in range(1, total_steps+1):
```

lona/client/_lona/rendering-engine.js
```diff
@@ -0,0 +1,637 @@
...
+            this._root.appendChild(node);
+
+        // HTML update
+        } else if(message_type == Lona.protocol.DATA_TYPE.HTML_UPDATE) {
+            this._widgets_to_setup = [];
+            this._widgets_to_update = [];
+
+            data.forEach(patch => {
+                this._apply_patch(patch);
+            });
```
maratori, line 630: nit:
```suggestion
            data.forEach(this._apply_patch);
```

lona/html/abstract_node.py
fscherf: What do you mean, the particular method or the entire feature? The feature gets used by the rendering test, to compare the server side node tree with the HTML that is shown in the browser.

tests/test_rendering.py
fscherf: I totally agree. Fixed.

tests/test_rendering.py
fscherf: I totally agree. Fixed.

lona/client/_lona/rendering-engine.js
fscherf: This won't work. The callback given to `data.forEach` has to be an arrow-function to preserve the scope of `this`. If written like this, this exception raises:

```
TypeError: Cannot read properties of undefined (reading '_apply_patch_to_node')                                                                          
    at _apply_patch (http://localhost:8080/static//_lona/rendering-engine.js:599:18)                                                                     
    at Array.forEach (<anonymous>)                                                                                                                       
    at LonaRenderingEngine.show_html (http://localhost:8080/static//_lona/rendering-engine.js:628:18)                                                    
    at LonaWindow._handle_websocket_message (http://localhost:8080/static//_lona/window.js:195:40)                                                       
    at LonaWindow.handle_websocket_message (http://localhost:8080/static//_lona/window.js:215:25)                                                        
    at WebSocket._handle_raw_websocket_message (http://localhost:8080/static//_lona/context.js:235:25) 
```

maratori: Wow, didn't know. Thanks!

lona/html/abstract_node.py
maratori: The question was about entire feature. Thanks for clarification.

maratori (approved):
LGTM

fscherf:
@maratori: Thank's for your feedback, and especially for digging into the JavaScript Code

PR #294, opened 2022-11-27
doc: make implications of awaiting input events and daemonizing clearer

fscherf:
Previously most examples in the documentation used the await API to retrieve input events. This API is meant for some special cases, like interactive, long-running wizards.
Also there was no clear distinction between daemonized- and multi-user views.

Users who are new to Lona, Python or Web in general (or all of them) might come to the conclusion that these are the APIs to use when creating a simple CRUD application.
This might work in simple test-cases, but will be very resource heavy with an actual user base.

This patch ports most examples to use callbacks to handle input events, and adds some clear warnings on the implications of the daemonize and the await APIs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,25 +53,20 @@ pythonic API to write self contained views.
 
     app = LonaApp(__file__)
 
+
     @app.route('/')
     class MyView(LonaView):
+        def handle_button_click(self, input_event):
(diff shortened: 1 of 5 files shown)


maratori (approved):
👍

fscherf:
@maratori: thanks for reviewing!

PR #301, opened 2022-12-14
html and input event fixes

fscherf:
This PR contains multiple fixes concerning the HTML API and input event API, and some refactoring

### DIFF
--- a/lona/client/_lona/context.js
+++ b/lona/client/_lona/context.js
@@ -53,7 +53,7 @@ export class LonaContext {
             this.settings.scroll_to_top_on_view_start = true;
         };
 
-        this._windows = {};
+        this._windows = new Map();
         this._last_window_id = 0;
         this._connect_hooks = [];
         this._disconnect_hooks = [];
@@ -72,11 +72,11 @@ export class LonaContext {
         this._last_window_id += 1;
 
         var window_id = this._last_window_id;
+        var lona_window = new LonaWindow(this, root, window_id, url);
 
-        this._windows[window_id] = new LonaWindow(
-            this, root, window_id, url);
+        this._windows.set(window_id, lona_window);
 
-        this._windows[window_id].setup(url);
+        lona_window.setup(url);
 
         return window_id;
     };
@@ -85,10 +85,10 @@ export class LonaContext {
     patch_input_events(root_node_selector, window_id) {
         // find window
         if(window_id == undefined) {
-            window_id = 1;
-        };
-
-        var _window = this._windows[window_id];
+            var _window = this.get_default_window();
+        } else {
+            var _window = this._windows.get(window_id);
+        }
 
         // patch input events
         var node = document.querySelector(root_node_selector);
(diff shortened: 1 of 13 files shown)


lona/html/node_list.py
```diff
@@ -181,16 +185,21 @@ def __contains__(self, node):
...
-    def _reset(self, value):
-        if not isinstance(value, list):
-            value = [value]
+    def _reset(self, values):
+        if not isinstance(values, list):
+            values = [values]
 
         with self._node.lock:
+            for node in self._nodes:
+                node._set_parent(None)
```
maratori, line 194: Looks like need to do the same in `__setitem__`

lona/client/_lona/context.js
```diff
@@ -53,7 +53,7 @@ export class LonaContext {
             this.settings.scroll_to_top_on_view_start = true;
         };
 
-        this._windows = {};
+        this._windows = new Map();
```
maratori, line 56: Looks line need to update `test_project/templates/lona/frontend.html` as well

lona/client/_lona/input-events.js
```diff
@@ -85,6 +85,7 @@ export class LonaInputEventHandler {
 
         node.onclick = function(event) {
             event.preventDefault();
+            event.stopPropagation();
```
maratori, line 88: This may be unexpected if someone uses custom js. Maybe need to add some comment to docs?

test_project/Makefile
```diff
@@ -4,32 +4,27 @@ PYTHON_VENV=env
...
 	. $(PYTHON_VENV)/bin/activate && \
 	pip install pip --upgrade && \
-	pip install -r ./REQUIREMENTS.txt && \
-	date > $(PYTHON_VENV)/.created
-
-env: $(PYTHON_VENV)/.created
+	pip install -r ./REQUIREMENTS.txt
 
-clean:
+clean: | $(PYTHON_VENV)
```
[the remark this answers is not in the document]
fscherf: Thanks for the question. I updated the commit message: https://github.com/lona-web-org/lona/pull/301/commits/4bcc27441b1b1c56047ccc24334193226001b4ac

lona/html/node_list.py
fscherf: Oh! You are right! Thanks for catching that. I added a patch that adds a similar change to `__setitem__`.

lona/client/_lona/context.js
fscherf: Thank's for catching that! I added a change to the patch for that.

lona/client/_lona/input-events.js
fscherf: You are right. I already noted that for the next release. I will add a note to https://lona-web.org/1.x/end-user-documentation/html.html#firing-custom-input-events in the release commit

maratori (approved):
LGTM

fscherf:
@maratori: Thanks a lot for your feedback and careful reading of this one!

PR #317, opened 2023-01-13
setup client 2

fscherf:
This PR contains the first version of the JavaScript client for Lona2. This code was initially merged into the `lona2` branch. That turned to be very unpractical, because both client implementations will need maintenance in the near future.
This PR makes the new client available behind a feature flag. This way we can easily develop the new implementation on the `master` and be fully compatible until we switch.

This PR also contains some fixes in components, i ran into issues, while splitting the JavaScript code.


fscherf:
The tests fail on Python 3.7, because ``shutil.copytree(dirs_exist_ok)`` was introduced in Python 3.8. I will need to find nother solution for collect-static, and update the patch before merging.

The rest of the patches should be ready for review

fscherf:
The collect-static problem is fixed on Python 3.7 now

PR #328, opened 2023-02-04
packaging: setup pyproject.toml

fscherf:
This patch sets up a `pyproject.toml` to fix multiple problems, with the current packaging, at once:

1. Deprecation warnings

The current setup, using a legacy `setup.py`, produces this warning since pip 23.0:

    DEPRECATION: lona is being installed using the legacy 'setup.py install'
    method, because it does not have a 'pyproject.toml' and the 'wheel' package
    is not installed. pip 23.1 will enforce this behaviour change. A possible
    replacement is to enable the '--use-pep517' option. Discussion can be found
    at https://github.com/pypa/pip/issues/8559

2. Problems with package data

On some systems, package data like the JavaScript client, or templates, are missing, when Lona is installed using the git URL:

    pip install git+https://github.com/lona-web-org/lona.git

This seems to be only reproducible on Windows.

3. `client2` was not packaged properly before

`client2` was not part of the `MANIFEST.in` before.

This patch gets rid of the old `setup.py`, and all related files like `MANIFEST.in`, and replaces all configuration with a `pyproject.toml` file.


sobolevn (commented):
Generally looks good! I have two questions:
1. Did you forget about adding classifiers from `setup.cfg`? Or is it intentional?
2. Why do you still need `REQUIREMENTS.packaging.txt`? You can add all requirements from there to `[build-system] requires = ["setuptools>=61.0", "twine", "build"]`

fscherf:
@sobolevn 
> 1. Did you forget about adding classifiers from `setup.cfg`? Or is it intentional?

That was not intentional. The whole file was more try-and-error for every line, and i removed classifiers by accident. I will add them to the `pyproject.toml`.

> 2. Why do you still need `REQUIREMENTS.packaging.txt`? You can add all requirements from there to `[build-system] requires = ["setuptools>=61.0", "twine", "build"]`

That's something i wanted to consult you about anyway: I want as few config files as possible, so i would want to merge all requirements files into the `pyproject.toml`, but than the `dependabot.yml` has to be updated right? Is this even supported/convenient in dependabot?

sobolevn:
> Is this even supported/convenient in dependabot?

No, I don't think so :(
But, you don't have to worry about that. Since all versions do not have exact pins.

fscherf:
> > Is this even supported/convenient in dependabot?
> 
> No, I don't think so :( But, you don't have to worry about that. Since all versions do not have exact pins.

That's right. Ok, so we merge `REQUIREMENTS.dev.txt` and `REQUIREMENTS.packaging.tx` into the `pyproject.toml` and leave `tests/REQUIREMENTS.test.txt` as is, for now?

sobolevn:
> Ok, so we merge REQUIREMENTS.dev.txt and REQUIREMENTS.packaging.tx into the pyproject.toml and leave tests/REQUIREMENTS.test.txt as is, for now?

I think that we should only merge `REQUIREMENTS.packaging.txt` and leave others.

fscherf:
> I think that we should only merge `REQUIREMENTS.packaging.txt` and leave others.

Why do you want to leave `REQUIREMENTS.dev.txt`?

sobolevn:
Because, as far as I know, `setuptools` does not support top-level dev-dependencies 🤔 
The only option is to use optional group like `.[dev]`

But, this polutes PyPI metadata and is not supported by dependabot.

Or am I missing something?

fscherf:
Makes sense!

PR #335, opened 2023-02-12
client: remove public class fields

fscherf:
This PR removes public class fields from both JavaScript clients, because this feature is not supported in Safari<14.1.
This produced issues on iPhone 6 and iPad mini 2, reported by @markusgritsch in #334.


markusgritsch:
Thanks a lot @fscherf for your fast fix! I can verify that it works fine on my iPhone 6 now.

fscherf:
@markusgritsch: I released 1.12.3, so you can use the fix in your project

PR #347, opened 2023-02-27
Fixed typos.

markusgritsch:


fscherf:
@markusgritsch: Great effort, thanks in advance! Could you please give your commit messages more content? For obvious fixes like these, an headline like `doc: end user documentation: getting started: fix typos` is enough

PR #348, opened 2023-02-27
doc: end user documentation: fix typos.

markusgritsch:


fscherf:
@markusgritsch: Thank you very much!

PR #353, opened 2023-03-05
doc: tutorial: getting started: update import string section

fscherf:
This section was a bit confusing, because import strings are only really relevant in Lona projects. Since the tutorial only uses scripts in its examples, it needed a bit of clarification, what import strings are meant for.


markusgritsch:
I think the intention behind this section is clear now.

fscherf:
@markusgritsch: Thanks for reviewing!

PR #365, opened 2023-03-16
client2: fix html parsing

fscherf:


fscherf:
The contents of this PR got moved to #368

PR #368, opened 2023-03-19
html and client fixes

fscherf:
This PR fixes multiple problems the HTML parsing API, the rendering code of the client, relative URL resolving of the client and introduces the new, more JavaScript like, widget API for the client.

@markusgritsch: This PR should fix both of your current issues.


markusgritsch:
> This PR should fix both of your current issues.

I can confirm that both issues are fixed.

markusgritsch (approved):
I can confirm that both issues are fixed.

fscherf:
@markusgritsch: Great! :) thanks for testing and reporting in the first place. I will release 1.12.4 this evening

PR #369, opened 2023-03-19
setup plotter demo

fscherf:
This PR adds a lona-chartjs demo showing a function plotter. This demo was originally contributed by @markusgritsch in #367


fscherf:
@markusgritsch: I created a commit as you for the original code, so the attribution does not get lost. I only did some basic changes, so the linter is happy. Then i ported the code to lona-picocss, and created a small clip of the demo.
Let me know whether you are ok with these changes.

markusgritsch:
I get this traceback:
```Python
  File "C:\Users\mgr\Desktop\lona-test\pl2.py", line 3, in <module>
    from lona_picocss.html import TextInput, Label, HTML, Grid, Sub, H1
ImportError: cannot import name 'Sub' from 'lona_picocss.html'```

markusgritsch:
Thanks for the honor of adding the example to Lona :)

markusgritsch (approved):
Works fine for me, and thanks for adding it.

PR #374, opened 2023-03-23
Bonsai.css fixes

fscherf:


fscherf:
This patch got moved to #380

PR #378, opened 2023-03-24
templates: frontend: Add viewport meta tag

markusgritsch:
Having this meta tag is recommended e.g. here and here and improves viewport size on mobile devices.


fscherf:
@markusgritsch: Nothing to complain about. Thanks for contributing! :)

PR #380, opened 2023-03-25
HTML refactoring

fscherf:
This PR adds all missing HTML5 nodes, adds support for node namespaces, to add support for SVG rendering, and adds support for non-standard CSS properties


markusgritsch (approved):
Great work! All my examples work fine with it.

fscherf:
This PR is finally done! I fixed the HTML parsing code, so it works with SVGs, and added `lona.html.RawHTML` which can render  raw HTML on the client without the server parsing it first.
For interactive SVGs that contain Javascript, we will need another solution, but with these patches we will have initial SVG support in Lona \o/

PR #394, opened 2023-04-09
doc: lona-scripts: fixed add_template() -> add_static_file()

markusgritsch:


fscherf:
Thanks for this patch!

Issue #401, opened 2023-04-17
Performance problems caused by NodeList


m-reichle:
We recently updated from lona 1.10 to 1.13 and noted that we are now experiencing performance problems when appending nodes. 
Running our test case through a profiler shows that the majority of time is spent on finding the position of newly added nodes at https://github.com/lona-web-org/lona/blob/master/lona/html/node_list.py#L68.
The index() method is quite expensive, since it runs an \_\_eq\_\_ (and thus two _serialize()) for each node in the list and the new node, every time a node is added, which makes runtime grow very quickly. 

```
               ncalls  tottime  percall  cumtime  percall    filename:lineno(function)
   10 nodes       45    0.000    0.000    0.004    0.000   abstract_node.py:49(__eq__)
  100 nodes    4'950    0.008    0.000    0.319    0.000   abstract_node.py:49(__eq__)
1'000 nodes  499'500    0.882    0.000   33.179    0.000   abstract_node.py:49(__eq__)
```
Collected via `cProfile.run('tr = lona.html.Tr()\nfor i in range(num): tr.append(lona.html.Td(i))', sort='ncalls')`
 
 Another thing we noticed during our testing is that since AbstractNode.\_\_eq\_\_() is called with include_node_ids=False, index() will actually not necessarily return the recently added node, but the first node with equal type and content.

Because of this 

    tr = lona.html.Tr()
    for i in range(num):
      tr.append(lona.html.Td('foo'))

will perform significantly better than 

    tr = lona.html.Tr()
    for i in range(num):
      tr.append(lona.html.Td(i))

This is probably a separate bug, though.

Looking at the code, calling index() seems unnecessary both in NodeList.append() and NodeList.insert() since in both cases the node's index is actually known.

The following patch would remove index() from these methods.

```
diff --git a/lona/html/node_list.py b/lona/html/node_list.py
index 90ea71afdc9f..48edcf456ae5 100644
--- a/lona/html/node_list.py
+++ b/lona/html/node_list.py
@@ -44,8 +44,6 @@ class NodeList:
 
             self._nodes.insert(index, node)
 
-            index = self._nodes.index(node)
-
             self._node.document.add_patch(
                 node_id=self._node.id,
                 patch_type=PATCH_TYPE.NODES,
@@ -57,25 +55,9 @@ class NodeList:
             )
 
     def append(self, value):
-        self._assert_not_frozen()
-        self._check_value(value)
-
         with self._node.lock:
-            node = self._prepare_node(value)
[...]

PR #402, opened 2023-04-21
Minimal fix for NodeList.append() and accompanying test

m-reichle:
Here's a minimal workaround for the performance problem caused by NodeList.append() and a small test to prove it's now performing normally. See #401 for details.

### DIFF
--- a/lona/html/node_list.py
+++ b/lona/html/node_list.py
@@ -65,7 +65,7 @@ class NodeList:
 
             self._nodes.append(node)
 
-            index = self._nodes.index(node)
+            index = len(self._nodes) - 1
 
             self._node.document.add_patch(
                 node_id=self._node.id,
--- a/tests/test_performance.py
+++ b/tests/test_performance.py
@@ -1,0 +1,14 @@
+from lona.html import Tr, Td
+
+
+def test_number_of_serialize_calls(mocker):
+
+    from lona.html import Node
+
+    spy = mocker.spy(Node, '_serialize')
+
+    tr = Tr()
+    for i in range(100):
+        tr.append(Td(i))
+
+    assert spy.call_count < 150
--- a/tests/REQUIREMENTS.test.txt
+++ b/tests/REQUIREMENTS.test.txt
@@ -8,5 +8,6 @@ coverage==7.2.3
 pytest==7.3.1
 pytest-aiohttp==1.0.4
 pytest-dependency==0.5.1
+pytest-mock==3.10.0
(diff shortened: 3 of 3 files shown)


lona/html/node_list.py
```diff
@@ -65,7 +65,7 @@ def append(self, value):
 
             self._nodes.append(node)
 
-            index = self._nodes.index(node)
+            index = len(self._nodes) - 1
```
fscherf, line 68: What if you append to an empty node, where the length is `0`?  The resulting index would be `-1` in this case.

tests/test_performance.py
```diff
@@ -0,0 +1,16 @@
+from lona.html import Tr, Td
+
+
+class TestPerformance:
```
fscherf, line 4: You wrote a unit-test style test. Could you port it to a pytest-style test (just a function, without the class)

tests/test_performance.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def test_number_of_serialize_calls(self, mocker):
+
+        from lona.html import Node
+
+        spy = mocker.spy(Node, "_serialize")
```
fscherf, line 10: Because of the double-quotes, the linter is unhappy here

lona/html/node_list.py
m-reichle: I can't find a scenario that would result in the node list still being empty at that point. Even if the node is empty initially, `len()` is called after appending, so there should always be at least the one node in the list. 

An `append(None)` could maybe get us there, but that's caught by `check_value()` and results in a `ValueError`.

fscherf: I am sorry, you are completely right. I overlooked that the `len()` always comes behind the append to `self._nodes`. So, a negative number is not possible.

tests/test_performance.py
m-reichle: Fixed in ec0315a43357b753d9ac2a57b0200bba0d6aee68

tests/test_performance.py
m-reichle: Fixed in ec0315a43357b753d9ac2a57b0200bba0d6aee68

tests/test_performance.py
m-reichle: Done in f71d682c99cd6b9048e8d49b4d90e8bfbb091669

fscherf:
@m-reichle: Great! I will merge as soon, as the CI is done. Thanks for contributing

fscherf (approved):

PR #414, opened 2023-05-29
tests: refactor docker setup

fscherf:
This PR improves the tox+docker setup for running the test-suite locally

The previous setup had multiple problems:

  1) **Tests ran as root in the container**
      All files, created by tox in the container belonged to root after the test-suite finished. That meant that caches and test artifacts, like screenshots or clips, had to be accessed using sudo.

  2) **Playwright installation steps could not be cached**
      The playwright installation commands, that setup the underlying ubuntu container and install the actual browsers, had to run on every run of the test-suite,  which is expensive.

  3) **The test container had only one, unspecified, Python version installed**
     The docker image for the test-suite is based on the Microsofts playwright image, which is based some Ubuntu image. So the Python version always was the current Python3 version on Ubuntu. This was problematic because the version could change on every run and testing with multiple Python versions was impossible.

This PR fixes all of these problems by adding a script that runs the test container as the current user, and updates the playwright docker file to install all Python versions that Lona supports, using pyenv.

@sobolevn: This PR also moves all requirement files into the `pyproject.toml` file. They were separate files before because GitHub Dependabot had no support for `pyproject.toml`. My understanding is that GitHub Dependabot added support for this ([source](https://github.blog/changelog/2022-10-24-dependabot-updates-support-for-the-python-pep-621-standard/)).
Could you take a look at my changes?


docker-compose
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# set the docker user to the local user so test environments and artifacts
+# can be access from inside and outside of the test container
+export DOCKER_USER=$(id -u):$(id -g)
```
sobolevn, line 5: This can be set via `docker-compose.yml`: https://github.com/wemake-services/wemake-django-template/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/docker-compose.override.yml#L23

playwright.Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM mcr.microsoft.com/playwright:v1.34.0-jammy
+
+# upgrade ubuntu
+RUN apt update && apt upgrade -y
+
+# setup tzdata (required by pyenv dependencies)
+RUN apt update && DEBIAN_FRONTEND=noninteractive TZ=Gmt/UTC apt-get -y install tzdata
+
+# install pyenv dependencies
+RUN apt update && apt install -y \
```
sobolevn, line 10: I would prefer to have one `RUN` command here. It feels more correct (I think that cache layers are not an issue anymore).

docker-compose
sobolevn: https://github.com/wemake-services/wemake-django-template/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/docker/django/Dockerfile#L13 yes, some extra tweeks are needed.

playwright.Dockerfile
fscherf: so `RUN apt update && apt upgrade -y && apt install $all_packages`?

I spitted these run commands because the last one is taken from here:  https://github.com/pyenv/pyenv/wiki
My intend was to have a section that can be updated easily when the wiki page changes

docker-compose
fscherf: Ah I see! But that only works if the GID and UID of the user that runs the test-suite are `1000` right?

fscherf: I added https://github.com/lona-web-org/lona/pull/414/commits/2b2dbddb13b88bbde461b21746639102f2ab3c92. You are right, it works great. Users that are not `1000:1000` on their system will have to edit `playwright.Dockerfile` and `docker-compose.yml` though. Maybe that's better than a magic script though.

playwright.Dockerfile
fscherf: good idea! Done!

sobolevn (approved):
Thank you!

fscherf:
@sobolevn: Thanks for the great feedback!

PR #432, opened 2023-07-03
python: drop support for 3.7

fscherf:
Since 27. Jun 2023, Python 3.7 is end of life (https://endoflife.date/python). Thus, this patch removes support for this version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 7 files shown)


sobolevn (approved):
You can also run `pyupgrade` if you wish (. It sometimes does good things to your code.

And don't forget to change `mypy`: https://github.com/lona-web-org/lona/blob/master/mypy.ini#L3

fscherf:
> You can also run `pyupgrade` if you wish (. It sometimes does good things to your code.

Nice tool! I ran it and it seems to be happy with the code.

> And don't forget to change `mypy`: https://github.com/lona-web-org/lona/blob/master/mypy.ini#L3

Good call. Fixed.

fscherf:
@sobolevn: Thanks for reviewing!

PR #434, opened 2023-07-06
refactor incremental tests

### LINKED ISSUE #433: AttributeError when comparing nodes
Hi,

the new version of `Abstract Node.__eq__()` has me run into errors with custom nodes that do not implement the attributes tested in `__eq__`. I'd say if the `AbstractNode` class has a method that assumes the existence of attributes, those should either have defaults or there should be a `hasattr()` beforehand. Not sure which would make more sense here.

I'd propose verifying this in a very basic test such as this.

```
from lona.html.abstract_node import AbstractNode

an1 = AbstractNode()
an2 = AbstractNode()

assert an1 == an1
assert an1 != an2
```

fscherf:
This PR refactors the incremental tests and fixes #433


m-reichle:
Thank you for taking care of this!
This PR fixes my issues described in #433 

Only noteworthy change I noticed is in a custom widget of mine (see below), where I just pass a single Node as `self.nodes`, this now results in an empty  `<NodeList([]))>` whereas before the NodeList would contain my icon. 

```
class CollapseIcon(Widget):

    def __init__(self, view):
        self.view = view
        self.html = view.html
        self.icon = I(
            _class="fas fa-minus-square",
            style={'cursor': 'pointer', 'color': '#3b8b5a', 'padding-right': '5px'},
            events=[CLICK]
        )
        self.nodes = self.icon
```

It is easily fixed doing a `self.nodes = [self.icon]` and I'm not sure if it was supposed to work before the change, but I thought I'd mention it since it's a change in behaviour.

lona/html/abstract_node.py
```diff
@@ -58,6 +58,11 @@ def __eq__(self, other):
         if self.NODE_TYPE == NODE_TYPE.TEXT_NODE:
             return self._string == other._string
 
+        # widgets
+        # TODO: remove in 2.0
```
sobolevn, line 62: Maybe issue a deprecation warning?

tests/test_0107_html_node_compare.py
```diff
@@ -0,0 +1,78 @@
...
+    assert TextNode('foo') is not TextNode('foo')
+
+    # legacy widgets
+    # TODO: remove in 2.0
+    assert TestWidget() == TestWidget()
+
+    assert TestWidget(Div()) == TestWidget(Div())
+    assert TestWidget(Div()) != TestWidget(Span())
+
+    assert TestWidget(foo=1) == TestWidget(foo=1)
```
sobolevn, line 28: I think that this test can be split into multiple ones.

fscherf:
@m-reichle:

> Only noteworthy change I noticed is in a custom widget of mine (see below), where I just pass a single Node as `self.nodes`, this now results in an empty `<NodeList([]))>` whereas before the NodeList would contain my icon.

Oh! Thanks for reporting, that is not supposed to happen. I updated the patch and added two tests so this may not happen again.

lona/html/abstract_node.py
fscherf: Until now we did not issue deprecation warnings but added wrote them to the documentation. It's a good idea to deprecate the Legacy Widget API now, but I think it's a little bit much for this PR because the old API gets used in multiple examples in the documentation. I opened #437 as a base for a follow-up PR.

tests/test_0107_html_node_compare.py
fscherf: Good idea. Done!

m-reichle:
> @m-reichle:
> 
> > Only noteworthy change I noticed is in a custom widget of mine (see below), where I just pass a single Node as `self.nodes`, this now results in an empty `<NodeList([]))>` whereas before the NodeList would contain my icon.
> 
> Oh! Thanks for reporting, that is not supposed to happen. I updated the patch and added two tests so this may not happen again.

Jup, runs without issues now!

fscherf:
@m-reichle: Great! Thanks for testing and reporting #433 in the first place!

Issue #229, opened 2022-08-09
Groups of radioboxes: Not all Python-Objects updated correctly?


SmithChart:
I am currently implementing an interface where I want to place a group of radioboxes (``<input type='radio' ...>``) where an user can select one-of-many.

I've started by copying the ``Checkbox`` from ``lona.html.data_binding_inputs`` to get the following:

```
class Radiobox(CheckBox):
    INPUT_ATTRIBUTE_NAME = 'checked'
    ATTRIBUTES = {
        'type': 'radio',
    }

    def __init__(self, value=False, disabled=False, readonly=False,
                 bubble_up=False, **kwargs):
        super().__init__(value, disabled, readonly, bubble_up, **kwargs)
        if 'input_delay' not in kwargs:
            self.events = [CHANGE]

    @property
    def value(self) -> bool:
        value = self.attributes.get(self.INPUT_ATTRIBUTE_NAME, False)
        if value == '':  # is possible after parsing HTML string
            return True
        return bool(value)

    @value.setter
    def value(self, new_value: bool) -> None:
        if not isinstance(new_value, bool):
            raise TypeError('value is a boolean property')
        self.attributes[self.INPUT_ATTRIBUTE_NAME] = new_value
```

Next step was to create a group of ``radioboxes`` like this:
```
Div(
    Radiobox(_id=1, name='groupname'),
    Radiobox(_id=2, name='groupname', checked=True),
)
```

When toggling the radiobox in the frontend to the other one I end up with both nodes on the python-side having ``checked=True``.
It seems as the other Radiobox does not emit an ``input_event`` to update it's state.

On a high level I would expect both Radoboxes to update at the same time.
What is the expected behavior here?

**Workaround:**
My plan was to have Radiogroup-Widget  anyway. So I have a single object that I can query for the current value and that creates the Radioboxes and Labels by itself.
But since I can not rely on the ``.value`` of the radio-boxes I am currently tracking the state of the group by hand. And this feels wrong.

Here is my prototype:
```
class Radiogroup(Widget):
    def __init__(self, options, checked=None):
        """
        Creates a Group of radiobuttons.

        Arguments:
        * options: Dict of options
                   { 'id-name 1': 'Screen Name 1', ... }
        * checked: String with the <id> of the pre-checked item
        """
        self._mapping = {}
        div = Div()
        for ident, screenname in options.items():
            rb = Radiobox(_id=ident, name=f'radiogroup-{self.id}', bubble_up=True)
[...]

PR #435, opened 2023-07-07
html: add radio buttons

fscherf:
This PR adds initial support for HTML radio buttons (#229)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 SHELL=/bin/bash
-PYTHON=python3.10
+PYTHON=python3.11
 PYTHON_ENV=env
 
 .PHONY: clean doc dist build test ci-test lint isort shell freeze
--- a/doc/content/api-reference/html.rst
+++ b/doc/content/api-reference/html.rst
@@ -827,6 +827,73 @@ select is no multi select, all other options get unselected automatically.
     |style             |(Dict) contains all styling attributes
 
 
+RadioGroup and RadioButton
+++++++++++++++++++++++++++
+
+.. code-block:: python
+
+    from lona.html import RadioGroup, RadioButton, Label
+
+    radio_group = RadioGroup(
+        Label('Option 1', RadioButton(value=1)),
+        Label('Option 2', RadioButton(value=2.0)),
+        Label('Option 3', RadioButton(value='3', checked=True)),
+    )
+
+    # adding radio buttons
+    # `RadioGroup.add_button()` takes any amount of nodes and connects
+    # the first `Label` with the first `RadioButton` object, using a
+    # random number and the HTML attribute `for`
+    radio_group.add_button(Label('Foo'), RadioButton(value='foo'))
+    radio_group.add_button(Div(Label('Foo'), RadioButton(value='foo')))
+
+    # if two non-node values are given, and the first one is a string,
+    # a `Label` and a `RadioButton` get inserted automatically
+    radio_group.add_button('Foo', 'foo')
+
+A ``RadioGroup`` consist of one or more ``RadioButton`` and ``Label`` object
+pairs, which hold information on value, checked state, and disabled state.
+
+``RadioButton`` objects consist of a value and a checked state. The value can
+be anything. If ``RadioButton.render_value`` is set, which is set by default,
+the content of ``RadioButton.value`` gets typecasted to a string and rendered
+into the HTML tree. This can be disabled if the actually values of the select
+shouldn't be disclosed to end users.
+
+``RadioGroup.value`` returns the value of the radio button that is currently
+checked.
+
+A radio button can be checked by setting ``RadioGroup.value`` to the value of
+the radio button that should be checked, or by setting ``RadioButton.checked``.
+
+**RadioGroup Attributes:**
+
+.. table::
+
+    ^Name                  ^Description
+    |bubble_up             |(Bool) Pass input events further
+    |radio_buttons         |(Tuple) tuple of all radio buttons
+    |checked_radio_button  |(Tuple) tuple of all selected options
+    |value                 |Value of the currently checked radio button
+    |values                |(Tuple) tuple of all possible values
+    |id_list               |(List) contains all ids
+    |class_list            |(List) contains all classes
+    |style                 |(Dict) contains all styling attributes
+
+**RadioButton Attributes:**
+
+.. table::
+
+    ^Name              ^Description
+    |name              |name of the radio button
+    |value             |value of the radio button
+    |checked           |(Bool) sets checked state
+    |disabled          |(Bool) sets the HTML attribute "disabled"
+    |id_list           |(List) contains all ids
+    |class_list        |(List) contains all classes
+    |style             |(Dict) contains all styling attributes
+
+
 Adding Javascript And CSS To HTML Nodes
 ---------------------------------------
 
--- a/lona/html/nodes/forms/radio_button.py
+++ b/lona/html/nodes/forms/radio_button.py
@@ -1,0 +1,198 @@
+from uuid import uuid1
+
+from lona.html.nodes.forms.inputs import TextInput
+from lona.html.nodes.text_content import Div
+from lona.html.nodes.forms import Label
+from lona.html.node import Node
+
+
+class RadioButton(TextInput):
+    INPUT_ATTRIBUTE_NAME = 'checked'
+
+    ATTRIBUTES = {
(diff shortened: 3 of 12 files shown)


lona/html/nodes/forms/radio_button.py
```diff
@@ -0,0 +1,136 @@
...
+    def checked(self, new_value):
+        if new_value:
+            self.attributes['checked'] = ''
+
+        else:
+            del self.attributes['checked']
+
+
+class RadioGroup(Node):
+    TAG_NAME = 'form'
```
SmithChart, line 70: The [HTML5 specification](https://html.spec.whatwg.org/multipage/forms.html#the-form-element) explicitly does not allow nesting of `<form>`. Even if it feels like there are not many use-cases where one want's to use actual HTML forms in Lona. Why make using them extra hard? 

Is there any gain from using `form` here instead of an element like `div` that is intended for nesting?

SmithChart (commented):
Nice to see some progress here :)

I like that even the `RadioGroup` feels like an HTML element in this context (even if HTML does not specify such node).

Do you see some room for a (specific) high-level interface like the one suggested [here](https://github.com/lona-web-org/lona/issues/229#issuecomment-1238281722) in Lona? Or should such interfaces live somewhere else? It feels wrong to have a custom widget in every project.

fscherf:
> Nice to see some progress here :)

Yeah, sorry :D

> Do you see some room for a (specific) high-level interface like the one suggested here in Lona? Or should such interfaces live somewhere else? It feels wrong to have a custom widget in every project.

I was thinking about something like `lona.html.RadioGroup.add_button()` too, but I am not sure which HTML layout such a method should produce. In picocss the default layout seems to be this:

```html
<label>
    My Label
    <input type="radio" value="my-value">
</label>
```

This seems to work fine in all cases I encountered but the "more HTML5" layout would be like this:

```html
<label for="my-radio-button>My Label</label>
<input type="radio" id="my-radio-button" value="my-value">
```

> [...] If I try to set a value that is not part of a RadioGroup I would suggest to raise an ValueError instead of silently not checking any RadioButton.

I agree. `lona.html.Select2` implements exactly this behavior so this should be consistent anyway.

> The HTML5 specification explicitly does not allow nesting of `<form>`. Even if it feels like there are not many use-cases where one want's to use actual HTML forms in Lona. Why make using them extra hard?
> 
> Is there any gain from using form here instead of an element like div that is intended for nesting?

Unfortunately, radio groups have no dedicated HTML tag but are identified by multiple radio buttons using the same name. When using a div as radio group, your code would have to look like this:

```python
RadioGroup(
    RadioButton(value='foo', name='my-group'),
    RadioButton(value='bar', name='my-group'),
    RadioButton(value='baz', name='my-group', checkd=True),
)```

If possible I don't want a mandatory name field that is not used by the application code since the `RadioGroup` object is the handle to the checked value. When using a `<form>` `RadioButton.name` can be generic, when using `<div>` not.

**Comment by SmithChart** [comment]
> I was thinking about something like `lona.html.RadioGroup.add_button()` too, but I am not sure which HTML layout such a method should produce. In picocss the default layout seems to be this:
> 
> ```
> <label>
>     My Label
>     <input type="radio" value="my-value">
> </label>
> ```
> 
> This seems to work fine in all cases I encountered but the "more HTML5" layout would be like this:
> 
> ```
> <label for="my-radio-button>My Label</label>
> <input type="radio" id="my-radio-button" value="my-value">
> ```

So, if both work, why care which one to use? (But I personally would prefer to stick closer to what HTML5 suggests.)
In my use-cases the choices for a RadioGroup often come from a (filtered) Django DB choice. Having `add_button()` would reduce boilerplate to:
```
choice = (("v1", "Display Value V2"), ("v2", "Display Value V2")) # already in the DB model

rg = RadioGroup()
[rg.add_button(v, k) for k,v in l]
```


> Unfortunately, radio groups have no dedicated HTML tag but are identified by multiple radio buttons using the same name. When using a div as radio group, your code would have to look like this:
> 
> ```python
> RadioGroup(
>     RadioButton(value='foo', name='my-group'),
>     RadioButton(value='bar', name='my-group'),
>     RadioButton(value='baz', name='my-group', checkd=True),
> )```
> 
> If possible I don't want a mandatory name field that is not used by the application code since the `RadioGroup` object is the handle to the checked value. When using a `<form>` `RadioButton.name` can be generic, when using `<div>` not.
> ```
OK, and when using a `form` instead of a `div` the default name is enough to group the RadioButtons, I see.
In this case using `form` is better than having some hidden magic in `RadioGroup` that takes care of the renaming.

**Comment by SmithChart** [comment]
> @SmithChart: I added an implementation for `RadioGroup.add_button()`. The method takes a variable list of nodes, searches for `Label` and `RadioButton` objects, and links them together, using the `for` attribute. So all of these calls work:
> 
> ```python
> radio_group.add_button(Label('foo'), RadioButton(value='foo'))
> radio_group.add_button(Div(Label('foo'), RadioButton(value='foo')))
> radio_group.add_button(Label('foo', RadioButton(value='foo')))
> ```
> 
> For convenience, I added a special case that makes code like this work too:
> 
> ```python
> radio_group.add_button('Foo', 'foo')
> ```
> 
> I am not sure if this is too much magic, but it seems to be useful. What do you think?

I get the point on magic arguments and flags you made for the Select.
For a RadioButton I think two magic arguments are still acceptable. This should be easily discoverable. And the _long form_ from above gives full flexibility.

So: I think I like this API.

**Comment by fscherf** [comment]
@SmithChart: Great! Thanks for your opinion on that. I will finish this PR, update the docs, and ping you for the final review

File: Makefile [hunk 2]
```diff
@@ -1,5 +1,5 @@
 SHELL=/bin/bash
-PYTHON=python3.10
+PYTHON=python3.11
```
SmithChart, line 2: Should this be part of this PR?

doc/content/api-reference/html.rst
```diff
@@ -827,6 +827,73 @@ select is no multi select, all other options get unselected automatically.
...
+++++++++++++++++++++++++++
+
+.. code-block:: python
+
+    from lona.html import RadioGroup, RadioButton, Label
+
+    radio_group = RadioGroup(
+        Label('Option 1', RadioButton(value=1)),
+        Label('Option 2', RadioButton(value=2.0)),
+        Label('Option 3', RadioButton(value='3'), checked=True),
```
SmithChart, line 840: `checked=True` is an argument for the `RadioButton`,  not the `Label`.

doc/content/api-reference/html.rst
```diff
@@ -827,6 +827,73 @@ select is no multi select, all other options get unselected automatically.
...
+    )
```
SmithChart, line 841: Initializing a RadioGroup via the constructor is not implemented! Using it this way Labels and RadioButtons are not linked together with the `for=` attribute.
Using `radio_group.add_button()` works however.

SmithChart (changes requested):
@fscherf I've reviewed your changes and gave it a try on a project I had open.

Makefile
fscherf: Not necessarily, but it isn't a problem either because it is a separate patch. While implementing this, I ran into some issues and wanted to use the new exception pretty printing of Python 3.11.

doc/content/api-reference/html.rst
fscherf: Good catch! Fixed.

doc/content/api-reference/html.rst
fscherf: That is due to the weird implementation of radio buttons in HTML. Correlating which `Label` is linked to which `RadioButton` is tricky when you get a whole tree of both. That's why, in this example, the `RadioButton` objects are inside of the `Label` objects. That way they are linked together by the browser implicitly. In `add_button()` that's not a problem, because you are supposed to provide only one label and one input with each call.

The `RadioGroup.value` is initialized correctly in this example because `value` is a property that searches for a `RadioButton` that is checked on every call. It is not a static value.

SmithChart: OK. Good point. Having the nodes in this way makes the result look funny. But `add_button()` works as expected. So no further action needed here.

SmithChart (approved):
LGTM

@fscherf feel free to close the resolved discussions. I can't do that from here.

fscherf:
@SmithChart: Great! Thanks a lot for your help on this topic, and for your patience. I know it took a long time.

Issue #268, opened 2022-10-19
RFC: Switch to disable String-to-HTML-parsing


SmithChart:
Cheers,
the real world hit my Lona project again. Before starting with a PR I would like to hear your opinion on how to go forward here.

**tl;dr:** I would like to add a settings-switch that disables the auto HTML-parsing in `HTML()` -objects.
IMHO auto-parsing is a footgun. With it enabled the developer has to escape each and every user input that could be passed (directly) into an HTML-object at some point. If the developers fails to do so a user can either break the application or inject custom HTML into the frontend.
And: With this feature enabled `HTML`-objects behave differently than other nodes (where there is no auto-parsing).

Consider the following example:
```
from lona.html import HTML, H1, Div
from lona import LonaApp, LonaView


app = LonaApp(__file__)


@app.route('/1')
class MyView(LonaView):
    def handle_request(self, request):
        html = HTML(
            H1('Hello World'),
            '<p>We can create HTML by simply writing some string here...</p>',
        )
        print(html.nodes)
        return html


@app.route('/2')
class MyView(LonaView):
    def handle_request(self, request):
        user_input = 'Somebody <somebody@example.com>'
        html = HTML(
            H1('Hello World'),
            user_input,
        )
        print(html)
        return html


@app.route('/3')
class MyView(LonaView):
    def handle_request(self, request):
        user_input = 'Somebody <somebody@example.com>'
        html = HTML(
            H1('Hello World'),
            Div(user_input),
        )
        print(html)
        return html


app.run(port=8080)

```

In `/1` Lona will take the String containing the `<p>` and will construct an actual `Node` from that. The output will be:
```
<h1 data-lona-node-id="1">
  Hello World
</h1>
<!--lona-widget:7-->
<p data-lona-node-id="5">
  We can create HTML by simply writing some string here...
</p>
<!--end-lona-widget:7-->
```

Now in `/2` we don't have a nice developer formatted string but some user input. (In our case this most often is an E-Mail message-ID we paste in a comment field for future reference).
Lona will assume that something containing `<` or `>` must be HTML and tries to parse it. This will fail:
```
LonaRuntimeWorker_0            ERROR    15:26:16.171305 lona.view_runtime Exception raised while running <__main__.MyView object at 0x7f9c782153d0>
  Traceback (most recent call last):
    File "/home/chris/work/Projects/intern/lag-intranet/env/lib/python3.9/site-packages/lona/view_runtime.py", line 318, in start
[...]

PR #436, opened 2023-07-07
html: add parse_html

fscherf:
This PR renames `lona.html.parsing.html_string_to_node_list` to `lona.html.parse_html` to make it official API that will replace HTML parsing in `lona.html.HTML` and `lona.html.HTML2` (#268)

### DIFF
--- a/doc/content/api-reference/html.rst
+++ b/doc/content/api-reference/html.rst
@@ -174,6 +174,40 @@ given selector.
 Using HTML Strings
 ~~~~~~~~~~~~~~~~~~
 
+To initialize an HTML tree you can use ``lona.html.parse_html``, which returns
+a Lona HTML node or a list of Lona HTML nodes.
+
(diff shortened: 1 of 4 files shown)


SmithChart:
How does renaming make this an official API?
We probably want to add it to the documentation though :-)

Otherwise LGTM.

fscherf:
> How does renaming make this an official API?

`lona.html.parsing.html_string_to_node_list` was never meant to be documented, because the name is so clunky. Giving it a cleaner name and putting it into the documentation makes it "official" :)

fscherf:
@SmithChart: I updated the documentation and all tests are green. I think in this state this is mergeable and closes #268

SmithChart (approved):
Thanks! Looks good.

fscherf:
@SmithChart: \o/ Thanks for your review and your patience

PR #438, opened 2023-07-10
fix static collection of deeply nested directories

fscherf:
This PR fixes static collection of deeply nested static files


tests/test_collect_static.py
```diff
@@ -221,3 +221,73 @@ def test_overlapping_directories():
...
     assert file1_content == 'tmp_dir1'
     assert file2_content == 'tmp_dir1'
+
+
+def test_deeply_nested_directories():
+    from lona.command_line.collect_static import collect_static
+
+    # setup temp dir
+    static_dir = TemporaryDirectory()
+    destination = TemporaryDirectory()
```
SmithChart, line 231: What about using `TemporaryDirectory` as context manager here? This way the directories will be cleaned up after the test has been completed. See https://docs.python.org/3/library/tempfile.html?highlight=temporarydirectory#tempfile-examples

fscherf: Good idea. I think for now it is ok because all tests are set up like this. So at least it is consistent. I will fix all tests in a follow-up PR

fscherf:
@SmithChart: Sorry, I was a bit fast with the merge-button this time. Thanks for your review anyway!

SmithChart:
@fscherf no worries. is mostly nitpicking anyway :-)

PR #442, opened 2023-07-13
fix rendering issue

fscherf:
This PR fixes a rendering problem in client2 when moving nodes within a tree, and contains a patch to make client errors more readable in log


SmithChart (approved):
I don't have any insight to the Javascript-side, so I am probably not the one to have an opinion on that.
Otherwise. LGTM.

fscherf:
@SmithChart: Thanks for having a look anyway!

PR #445, opened 2023-07-24
remove unused method

AvdN:
Index.handle_click is never used in this example, looks like it is a remnant of copy-paste of example-1.py


fscherf:
You are right, and that's pretty certainly how it happened :) thanks for fixing it!

PR #446, opened 2023-07-25
Deprecated example code, clarification

AvdN:
- the non-interactive example returning JSON should not use the deprecated dictionary return
- a line of explanatory text for the Trailing Slash example (had to test if this actually worked as I suspected).


AvdN:
Let me know if you prefer to have these kind of unrelated changes in two PRs, for docs I tend to go by file.

fscherf:
Nice! Thanks for that.
No that's no problem, especially not for documentation, since the changes should be self-explanatory anyway.

PR #453, opened 2023-07-27
views: warn if handle_request returns dict

AvdN:
With this change a warnings.warn() is invoked when a handle_request returns a dict instead of a response objec. This warning is issued once for each unique file-name, line-number of the "offending" handle_request(s). The warning references the file and line number of the "offending" handle_request method and a link to the Response Objects documentation.

The end-user can influence the result of the warning in the usual way by including the following lines:

   import warnings
   from lona.warnings import DictResponseDeprecationWarning

   warnings.simplefilter('always', DictResponseDeprecationWarning)

causing the warning to issue every time (other options include 'error' and 'ignore')

The warnings.warn mechanism is patched to be able to handle printing callee file-name and line-number (instead of the more usual caller info), in a similar way as the package ruamel.std.warnings.

### DIFF
--- a/lona/view_runtime.py
+++ b/lona/view_runtime.py
@@ -350,7 +350,7 @@ class ViewRuntime:
                 lona.warnings.warn(  # NOQA: G010
                     'Deprecated use of dict as return value of method handle_request\n  (see: https://lona-web.org/1.x/api-reference/views.html#response-objects)',
                     lona.warnings.DictResponseDeprecationWarning,
-                    callee=self.view.handle_request
+                    callee=self.view.handle_request,
                 )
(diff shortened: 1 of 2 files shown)


AvdN:
Let me know if you cannot fold the multiple commits into one, than I'll redo the PR (once more)

fscherf (changes requested):
Nicely done! Just some minor nitpicking from my side, but nothing major.
Please squash all commits into one. You don't have to start a new PR. Just force-push to this branch after squashing.

AvdN:
Now it doesn't start the automated check anymore...

PR #454, opened 2023-07-27
views: warn if handle_request returns dict

AvdN:
With this change a warnings.warn() is invoked when a handle_request returns a dict instead of a response objec. This warning is issued once for each unique file-name, line-number of the "offending" handle_request(s). The warning references the file and line number of the "offending" handle_request method and a link to the Response Objects documentation.

The end-user can influence the result of the warning in the usual way by including the following lines:

   import warnings
   from lona.warnings import DictResponseDeprecationWarning

   warnings.simplefilter('always', DictResponseDeprecationWarning)

causing the warning to issue every time (other options include 'error' and 'ignore')

The warnings.warn mechanism is patched to be able to handle printing callee file-name and line-number (instead of the more usual caller info), in a similar way as the package ruamel.std.warnings.

### DIFF
--- a/lona/view_runtime.py
+++ b/lona/view_runtime.py
@@ -42,6 +42,7 @@ from lona.html.document import Document
 from lona.connection import Connection
 from lona.request import Request
 from lona.routing import Route
+import lona.warnings
 
 # avoid import cycles
 if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


fscherf:
@AvdN: Really nice! I did test it locally and it seems to have one issue: The warning also gets printed when `handle_request` returns a string, which is a supported use-case. That might be a problem in the response parser though. I will investigate this further.

fscherf:
@AvdN: Turns out: the default frontend view in lona returns a dict response. So your code is working better than we thought :)
Thanks for contributing, and for this cool feature! Next thing could be coloring warnings yellow in the shell so it matches the warnings in the logging setup

PR #458, opened 2023-07-29
views: issue warning on use of .daemonize

AvdN:
add deprecation warning
add intermediary Lona_2_0_DeprecationWarning so only one warnings needs to be set for filtering.
update example code

### DIFF
--- a/doc/content/demos/daemonized-view/daemonized-view.py
+++ b/doc/content/demos/daemonized-view/daemonized-view.py
@@ -46,7 +46,7 @@ class DaemonView(LonaView):
         )
 
         # tell Lona to not kill the view when the page gets refreshed
-        self.daemonize()
+        self.is_daemon = True
 
         while True:
             self.show(html)
--- a/lona/html/node.py
+++ b/lona/html/node.py
@@ -13,6 +13,7 @@ from lona.html.node_list import NodeList
 from lona.protocol import NODE_TYPE
 from lona.html.widget import Widget
 from lona.state import State
+import lona.warnings
 
 
 def parse_style_string(style_string: str) -> dict[str, str]:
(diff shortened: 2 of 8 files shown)


SmithChart:
No need to change this for this PR, but shouldn't we issue deprecation warning at all the other soon-to-be-deprecated features, too?

fscherf:
@SmithChart, @AvdN: I created #468 with a roadmap. The PR still needs review, but I think you can use the list of tasks for Lona 2 to create warnings and deprecations

lona/compat.py
```diff
@@ -14,6 +16,7 @@
 def get_client_version():
     # TODO: remove in 2.0
 
+    lona.warnings.remove_2_0()
```
fscherf, line 19: All warnings in this module are incorrect because they run no matter what the user does, and are meant to remain compatible with old API. The `TODO: remove in 2.0` is only there so we don't forget to remove the functions before releasing Lona 2

lona/view_runtime.py
```diff
@@ -123,6 +123,8 @@ def __init__(
 
         # TODO: remove in 2.0
         # compatibility for older Lona application code
+        if hasattr(self.view, 'STOP_DAEMON_WHEN_VIEW_FINISHES'):
```
fscherf, line 126: This is a feature flag. Not using it should not result in a deprecation warning

fscherf (changes requested):
Really nice! I like the approach. Needs only a little bit of work.

fscherf:
@AvdN: This PR is open for a long time now. Should I just take over and add the last fixes?

AvdN:
Hi @fscherf, yes please, I am not sure why I didn't react to this at the first request. Sorry for that.

fscherf:
@AvdN: There is nothing to be sorry about :)

fscherf:
@AvdN: merged! I am so sorry that it took me so long. The test suite was broken for a long time and it took me a long time to understand why. Thanks for contributing, and for your patience!

PR #462, opened 2023-08-07
client1 + client2: Fix `onsubmit` event handler to allow submitting forms

SmithChart:
According to the documentation
(https://lona-web.org/1.x/api-reference/views.html#post ) it should be possible to submit a `<form>` using `POST` to an interactive lona view. Without this change this fails with a
`Uncaught ReferenceError: lona_window is not defined` in the browser.

This change fixes the missing reference and allows to use the `POST` method to submit data to the view.

### DIFF
--- a/lona/client/_lona/client/input-events.js
+++ b/lona/client/_lona/client/input-events.js
@@ -202,6 +202,8 @@ export class LonaInputEventHandler {
     };
 
     _patch_onsubmit(node) {
+        var lona_window = this.lona_window;
+
         node.onsubmit = function(event) {
             event.preventDefault();
             event.stopPropagation();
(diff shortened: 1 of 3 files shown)


fscherf:
@SmithChart: Nice! Can you add a test too?

SmithChart:
Since tests rely on Docker I am not sure if I can actually run the tests anywhere. I'll look into that.

SmithChart:
@fscherf please have another look. I've added a test. But I am not sure if it should be put somewhere specific into the ordered block. From my point of this looks OK :-)

fscherf:
@SmithChart: Nice! I refactored your test a bit (https://github.com/lona-web-org/lona/tree/fscherf/fix-post-requests)
If you ok with the changes just squash my commit into yours. Also please remove the black configuration from the pyproject.toml.
After that the PR should be mergable

SmithChart:
@fscherf: I have squashed your suggested changes. Thanks for the update!

fscherf:
@SmithChart: Great! Thanks for contributing!

PR #464, opened 2023-08-07
server: get default port optionally form environment var

AvdN:
If the default port 8080 is taken, you have to either change the app.run() or  use --port <PORTNUM> on every invocation of the program. When trying out examples I don't want to do the first and forget the second.

This patch allows you set the environment variables LONA_DEFAULT_HOST and LONA_DEFAULT_PORT in your shell. These override the built-in defaults ('localhost' resp. 8080), and are in turn overridden by arguments to ``app.run()`` and startup commandline parameters.

### DIFF
--- a/lona/app.py
+++ b/lona/app.py
@@ -299,12 +299,14 @@ class App:
         from lona.command_line.handle_command_line import (
             parse_overrides,
             DESCRIPTION,
+            EPILOG,
         )
 
         parser = ArgumentParser(
             prog=str(self.script_path),
             formatter_class=RawTextHelpFormatter,
             description=DESCRIPTION,
+            epilog=EPILOG,
         )
 
         parser.add_argument(
@@ -428,8 +430,8 @@ class App:
 
         # setup arguments
         server_args = Namespace(
-            host='localhost',
-            port=8080,
+            host=os.environ.get('LONA_DEFAULT_HOST', 'localhost'),
+            port=os.environ.get('LONA_DEFAULT_PORT', '8080'),
             shell_server_url='',
             shutdown_timeout=0,
             log_level='info',
--- a/doc/content/tutorial/01-getting-started/index.rst
+++ b/doc/content/tutorial/01-getting-started/index.rst
@@ -101,6 +101,9 @@ Run the script using:
 
 ``app.run()`` takes keywords like ``host`` or ``port`` for configuration, and
 also parses the command line. Run ``python example.py -h`` to print the help.
+If port 8080 is taken by another application, you can set environment
+variable ``LONA_DEFAULT_PORT`` to some other port, affecting all examples that 
+do not set the port explicitly.
 
 The script should print that it opened a webserver on
 ``http://localhost:8080``. If you navigate your browser there, you should see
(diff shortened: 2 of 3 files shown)


lona/app.py
```diff
@@ -428,8 +430,8 @@ def run(
 
         # setup arguments
         server_args = Namespace(
-            host='localhost',
-            port=8080,
+            host=os.environ.get('LONA_DEFAULT_HOST', 'localhost'),
+            port=int(os.environ.get('LONA_DEFAULT_PORT', 8080)),
```
fscherf, line 434: We will need a check here if the value of `LONA_DEFAULT_PORT` can be type casted to an int. This will crash into a cryptic traceback if not

fscherf (changes requested):
Really nice! Thanks for this feature. Only a very small change is required to merge it.
Could you add the prefix "server: " to the commit message as well?

AvdN:
had not realised that if you amend the title of the commit, that doesn't percolate through into the PR.
If you do
```
 LONA_DEFAULT_PORT=abc python example.py
```
you'll get a message:
```
value "abc" for environment variable "LONA_DEFAULT_PORT" cannot be converted to an integer port number
```
after which the system sys.exits with value 2 (same value as when using `--port abc`, however I realise you cannot have the above and specify --port 8081, as it will still error on the environment variable not converting. This might be circumvented by assigning the environment variable to port and doing the check-for-int-ness after ``setattr`` has optionally been updated by the commandline-parsing.

But if you have an incorrect LONA_DEFAULT_PORT setting, IMO that should be fixed, not suppressed by doing --port.

fscherf:
I did some testing and found out the check is just not necessary (sorry). Argparse does exactly what you would expect and will handle `8080` and `"8080"`. So we just have to add `port=env.get('LONA_DEFAULT_PORT', '8080')` in these two spots, and everything should be fine.

https://github.com/lona-web-org/lona/blob/7f8ba4c28aa25453cf384707b80f33765316c744/lona/app.py#L432
https://github.com/lona-web-org/lona/blob/7f8ba4c28aa25453cf384707b80f33765316c744/lona/command_line/handle_command_line.py#L119

AvdN:
> I did some testing and found out the check is just not necessary (sorry). Argparse does exactly what you would expect and will handle `8080` and `"8080"`. So we just have to add `port=env.get('LONA_DEFAULT_PORT', '8080')` in these two spots, and everything should be fine.
> 
> https://github.com/lona-web-org/lona/blob/7f8ba4c28aa25453cf384707b80f33765316c744/lona/app.py#L432
> 
> https://github.com/lona-web-org/lona/blob/7f8ba4c28aa25453cf384707b80f33765316c744/lona/command_line/handle_command_line.py#L119

I did this but cannot re-apply for a review, probably because there is already a review request pending.

fscherf:
@AvdN: Yeah no worries, that's on me.
I think that's it! I will do some more local testing, but the patch should be mergeable in this state

fscherf:
@AvdN: Works like a charm! I added a small fix to your patch because `LONA_DEFAULT_HOST` did only work for apps, not for projects. Let me know if you are ok with this change, then we can merge.

fscherf (approved):

AvdN:
I am not sure if I am supposed to be able to see the changes on github, at least I cannot find them. But go ahead, I'll see them at the latest when they are merged.

fscherf:
@AvdN: Thanks for contributing! :)

PR #552, opened 2024-03-18
fix browser history

### LINKED ISSUE #550: Browser back button not working in Firefox
Hi, 

I'm currently looking into an issue where the browser's back button does not work as expected when accessing Lona views using the Firefox browser. It works as expected when using Chrome/-ium but only seems to create a reload in Firefox.

Has anyone observed the same problem and/or found a way to solve it? I'm mostly puzzled by this only occurring in one browser.

fscherf:
Fixes #550


rschwebel:
Is it intended that the patches still have a WIP: subject?

rschwebel (commented):
Could you add some reasoning why the playwright version needs to be downgraded? Makes it easier in the future to find out if the reason is still valid.

fscherf:
Hi @rschwebel! As you correctly pointed out, the patches are not cleaned up yet and won't be merged in this state. I wanted some early feedback from @m-reichle if my changes fix her problem (or I suppose yours, for that matter :) ). Also I wanted to check wether the test works in CI.

While working on this, I had a lot of trouble running the browser tests locally. All calls to Playwrights `page.wait_for_url('/')` timeout for me and result in an InvalidStateError. It seems to be a regression because downgrading helped.
I am not sure yet what to do here and if this patch will on this PR. I will try to debug it further and report my findings to Playwright upstream if necessary.

rschwebel (commented):
Looks good, thx!

Issue #632, opened 2026-01-07
python3.14 break command_line/run_server.py


pfpulux:
Hi the update to python3.14 break the run_server.py with

```
  File "/home/void/.local/lib/python3.14/site-packages/lona/app.py", line 437, in _run
    run_server(
    ~~~~~~~~~~^
        args=server_args,
        ^^^^^^^^^^^^^^^^^
        server=self.server,
        ^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/void/.local/lib/python3.14/site-packages/lona/command_line/run_server.py", line 23, in run_server
    loop = asyncio.get_event_loop()
  File "/usr/lib/python3.14/asyncio/events.py", line 715, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
                       % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

See also the python3.14 [Changelog](https://docs.python.org/3/library/asyncio-eventloop.html).

I looks like `sed -i "s/get_event_loop/new_event_loop/" run_server.py` will fix this, but I'm not have checkt for side-efects.

fscherf:
Thanks for reporting! We need to maintain compatibility with older versions though. I'll add a fix and a python 3.14  environment to the test suite.

janheini:
We faced the same issue with Flamingo and it turned out to be a one line fix:
https://github.com/pengutronix/flamingo/pull/57

I haven't looked at the code too closely but maybe the fix here is similar.

fscherf:
Currently we support for python versions up to 3.11. I looked into adding support up to 3.14 yesterday, and there are multiple smaller issues. No big road block as far as I can see so far. Working on it.

@janheini will be the same story for flamingo I assume. Both projects were conceptualized around the same time.

fscherf:
@pfpulux can you please verify that #636 fixes your problem?

PR #634, opened 2026-04-22
server: reverse: fix kwargs type annotation

janheini:
Hey,

when annotating kwargs the type annotation refers to the type of the items within the dict [^1]. This is already done correctly around the codebase, except for `Server.reverse`.

Let's drop the dict there as well so my language server will stop yelling at me :)

Thanks!

[^1]: https://typing.python.org/en/latest/spec/callables.html#annotating-args-and-kwargs

### DIFF
--- a/lona/server.py
+++ b/lona/server.py
@@ -990,7 +990,7 @@ class Server:
             self,
             route_name: str,
             *args: Any,
-            **kwargs: dict[str, Any],
+            **kwargs: Any,
     ) -> str:
 
         """


fscherf:
Hi @janheini,

Makes sense. Do you need a release for this fix?

Thx for contributing

janheini:
Thanks for merging!

> Makes sense. Do you need a release for this fix?

Not necessarily. You will probably add a new point release when the Python 3.14 compatibility (#632) is fixed. I'm fine with waiting until then :)

fscherf:
@janheini I'm also trying to get @SmithChart open PR into the next release as well.

